PR #334, opened 2023-03-29
Stefankracht/dir 211 add chromatic to ci

stefan-kracht:


stefan-kracht:
I resolved this by setting up a new vite project with storybook 7. I was able to publish this to chromatic with no problems. Then I looked how the configuration differed from our setup and found out about the `.storybook/preview-head.html` file that defined `global`.

PR #349, opened 2023-04-12
DIR-85 Tabs component

porleaaron:
Basic UI and stories for the Tabs component


stefan-kracht (changes requested):
[See ticked comments](https://linear.app/direktiv/issue/DIR-85/tabs)

PR #354, opened 2023-04-25
[DIR-213] migrate away from daisyui

porleaaron:


stefan-kracht (changes requested):
I left some comments and tickets in linear with some change request. You can go through all of them and reopen this PR after you are done with 👍

PR #355, opened 2023-04-26
[DIR-433] Command Component

porleaaron:


stefan-kracht (changes requested):
Looks good. I just have some [design change requests](https://linear.app/direktiv/issue/DIR-499/update-styling).

PR #367, opened 2023-05-10
draft: [DIR-548] webkit workaround

sebxian:
Please do not merge, we should discuss this first.

See the comments in the file.

This will "force" clicks only in the context where this is needed (when tests run locally on webkit).

However, I am not really comfortable with the logical complexity this adds to the tests. I think I'd rather keep them simple and live with tests failing locally in webkit until we know more.


sebxian:
Closed, we decided against implementing this as it would add too much complexity to the tests.

PR #370, opened 2023-05-11
[DIR-484] Workaround to prevent breadcrumb formatting

sebxian:
A relatively crude workaround, but it may be good as a temporary fix until we upgrade to the new UI.

use-react-router-breadcrumbs seems to use the "humanize" method which introduces unwanted formatting, [unless a custom name for the route is provided](https://github.com/icd2k3/use-react-router-breadcrumbs/blob/52a93561a817d6aba7362e61b2f2a5fbf00fa230/src/index.tsx#L362).

So this workaround extracts our url segments and feeds them to useBreadcrumbs as "custom" names.

@stefan-kracht I couldn't get the segments via the useParams() hook (it just returns an empty object), not sure if there's a better way to get good segment names?

To do (maybe): Do we want breadcrumbs for "Explorer" etc. to be capitalized? That should be possible by just adding more custom route objects to the array.

### DIFF
--- a/src/components/breadcrumbs/index.tsx
+++ b/src/components/breadcrumbs/index.tsx
@@ -1,18 +1,11 @@
 import "./style.css";
 
-import { Link, useSearchParams } from "react-router-dom";
+import { Link, useLocation, useSearchParams } from "react-router-dom";
 
 import FlexBox from "../flexbox";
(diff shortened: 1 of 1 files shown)


stefan-kracht:
Since the next release will already be ahead of this issue, I will close that for now

PR #374, opened 2023-05-17
[DIR-527] use react i18next in all new components

porleaaron:


stefan-kracht:
@porleaaron can you check the failing tests?

PR #375, opened 2023-05-19
Stefankracht/dir 571 add traffic shapping

stefan-kracht:


src/api/tree/mutate/deleteNode.ts
```diff
@@ -12,6 +12,27 @@ import { useApiKey } from "~/util/store/apiKey";
 import { useNamespace } from "~/util/store/namespace";
 import { useToast } from "~/design/Toast";
 
+const updateCache = (
```
stefan-kracht, line 15: I creared a `updateCache` method here as well. This one has a second parameter, which is a bit more trickey to type manually.

e2e/explorer/workflow/revisions/trafficShaping.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+      name,
+    },
+  });
+
+  await page.goto(`/${namespace}/explorer/workflow/revisions/${name}`);
+  await expect(
+    page.getByTestId("traffic-shaping-container"),
+    "it renders the traffic shaping component"
+  ).toBeVisible();
+
```
stefan-kracht, line 37: Some general thoughts where I'm interested in your feedback:

In some places I use hard coded texts from the translation files and I wonder if we can keep it that way. Generally I would prefer to use a test id, but I have some cases where it seems to make sense. The downside is, that the tests break when we change some translations, but it's also a very easy fix.

Examples where I use text in this file:

"Select Revision..." this is the "best" way to find out if the dropdown is unused. I could add a helper attribute to the button, but then I would just test the data attribute.

"Please select 2 different revisions to configure traffic shaping." - I would prefer this, instead of introducing a redundant DOM element

"The traffic will be split between NAMEA and NAMEB with a ratio of X to Y %" - I want to be very sure that the user sees this text and also that we fill the placeholder from the translation file correctly

stefan-kracht:
I noticed that the e2e folder was not yet included in our tsconfig, which made it run with the default TS settings. I changed that to make it consistent across the project and had to add some types to the utils ("implicit any" was allowed before).

I will need some practice for e2e testing but I generally like it a lot. It just randomly fails in CI until I rerun. Don't know why.

I tried the api functions from our api folder instead of rebuilding the requests, just to try it out. I think I like this approach (I tried the manual approach too and it feels like I have to rewrite code that I already wrote weeks ago :D), just some thoughts that came into my mind:

**PROS:** 
  - faster to write e2e tests
  - types for free
  - will be very convenient when the API layout will change
  - when we would reimplement all requests, we need to type the responses (TypeScript now complains about no implicit any), so we either have to type the responses manually or import the typescript from the schema files

**CONS:**
  - no separation of concerns anymore, we need to import application code into our e2e test folder (but this might also be required to solve TS issues when rebuilding the api requests manually)
  - our tests are not that unbiased anymore. Having application code in the e2e tests, we "theoretically" could introduce bugs in the app and in the tests as well.

sebxian:
> 

I agree this looks quite good in the files. The cons are something to keep in mind, but for now, I think the pros may outweigh the cons.

e2e/explorer/workflow/revisions/trafficShaping.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+  await expect(
+    secondRevisionRow.getByTestId("traffic-distribution-secondary")
+  ).toHaveText(`${weightB} % of traffic distribution`);
+
+  // reload page
+  await page.reload();
+
+  await expect(
+    page.getByTestId("route-a-selector"),
+    "after reloasing, route a selector is prefilled with the previously selected revision"
```
sebxian, line 166:
```suggestion
    "after reloading, route a selector is prefilled with the previously selected revision"
```

e2e/explorer/workflow/revisions/trafficShaping.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+  ).toHaveText(revisionA.slice(0, 8));
+
+  await expect(
+    page.getByTestId("route-b-selector"),
+    "after reloasing, route b selector is prefilled with the previously selected revision"
```
sebxian, line 171:
```suggestion
    "after reloading, route b selector is prefilled with the previously selected revision"
```

sebxian (changes requested):
Found a typo, otherwise all is good :-)

sebxian (approved):

PR #376, opened 2023-05-19
[DIR-559] Active Revision Page e2e tests

porleaaron:


e2e/explorer/workflow/active.spec.ts
```diff
@@ -1 +1,138 @@
...
+
+let namespace = "";
+let workflow = "";
+const defaultDescription = 'A simple \'no-op\' state that returns \'Hello world!\''
+
+test.beforeEach(async ({page}) => {
+  namespace = await createNamespace();
+  workflow = await createWorkflow(namespace, faker.git.shortSha() + '.yaml');
+  test.setTimeout(120000)
+  await page.goto("/");
```
sebxian, line 20: It is good to have one test that navigates to the revision from the namespace. But I think it's not ideal to repeat all these steps in the beforeTest() hook. When we add more tests, the time that this takes will add up. I think it's better to keep the beforeTest() short, it should focus on setup and not include too many assertions.

I think it would be better to have one test for only "it is possible to navigate to the active revision", then every test for specific functions can start by visiting `${namespace}/explorer/workflow/active/${workflow}`.

e2e/explorer/workflow/active.spec.ts
```diff
@@ -1 +1,138 @@
...
+
+  await expect(page, "the namespace is reflected in the url").toHaveURL(
+    `/${namespace}/explorer/tree`
+  );
+  await expect(
+    page.getByTestId("breadcrumb-namespace"),
+    "the namespace is reflected in the breadcrumbs"
+  ).toHaveText(namespace);
+
+  await page.getByTestId(`explorer-item-link-${workflow}`).click()
```
sebxian, line 45: After navigating to the revisions page, it would be good to check that the breadcrumb for the workflow exists.

e2e/explorer/workflow/active.spec.ts
```diff
@@ -1 +1,138 @@
...
+  const toastAction = page.getByTestId("make-revision-toast-success-action");
+  await expect(toastAction, 'success toast should appear after make-revision button click').toBeVisible();
+
+  await toastAction.click();
+  await expect(toastAction, 'success toast should disappear after toast action click').toBeHidden();
+  await expect(page, "page url should have revision param").toHaveURL(/revision=/);
+  const revisionId = page.url().split('revision=')[1];
+  await expect(page.getByText(revisionId), "revisionId should be in the revision list").toBeVisible();
+  //go back to the workflow editor
+  await page.getByTestId('workflow-tabs-trg-activeRevision').click()
```
sebxian, line 94: A test should not end with an action, we should always assert that it has the intended result.

e2e/explorer/workflow/active.spec.ts
```diff
@@ -1 +1,138 @@
...
+}
+
+test("it is possible to save the workflow", async ({
+  page,
+}) => {
+  // click on the description so it can have input focus
```
sebxian, line 100: This comment is confusing, I would remove it or move it into the testSaveWorkflow method

e2e/explorer/workflow/active.spec.ts
```diff
@@ -1 +1,138 @@
...
+  await testSaveWorkflow(page)
+});
+
+test("it is possible to make the revision", async ({
+  page,
+}) => {
+  //now click on Revision Menu Trigger
```
sebxian, line 107: See above, remove or move to the correct context in the code

sebxian (commented):
Good work! Very good structure overall and I like the very clear naming of the testids. In my comments, you'll find some thoughts on how to improve the tests. Since we are just getting started developing our "best practices" for testing, these are open to discussion, so let me know what you think about my suggestions.

I added one commit with mainly wording changes (see my comments).

Question: Did you talk to Stefan about using Prettier? I think both of us use it with the setting `format on save` enabled in VSCode. Your spec file doesn't seem to be formatted by Prettier.

We can decide about this when Stefan is back, but I think it would be good if everybody uses the same setup, to avoid formatting related diffs in git (which just make the diffs hard to read). For my small commit, I disabled auto-format for the spec file so it is easier to see for you what I have changed.

PR #378, opened 2023-05-26
[DIR-569] JSONSchemaForm basics

porleaaron:


stefan-kracht (changes requested):
This looks very good Aaron. I'm happy you figured that out.

Regarding the `.vscode/settings.json`, I changed it so that the file is deleted from our repo but added to `.gitignore` so that every one of us can just setup the editor as we like.

I have some suggestions on how we can move on with this before we merge:

- I would love to have see some more form primitives. Can you update the JSONschema in the storybook to show a 
  - select
  - a checkbox
  - a radio button
- minor design stuff
  - Can you move the button submit button to the right?
  - a little bit more spacing between the "rows" of the form would be nice.
  - the font size of the label and the headline should be the same as in our overlays in the app (but I like that you have the labels on top of the input and not next to it, it makes it easier to style)  Bildschirm­foto 2023-05-31 um 11 08 32

- fix all open TS issues

src/design/JSONschemaForm/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+  SelectValue,
+} from "../Select";
+
+import Button from "../Button";
+import { Checkbox } from "../Checkbox";
+import Form from "@rjsf/core";
+import Input from "../Input";
+import validator from "@rjsf/validator-ajv8";
+
+const CustomSelectWidget: React.FunctionComponent<WidgetProps> = (props) => (
```
stefan-kracht, line 35: can we use FC instead of React.FunctionComponent

src/design/JSONschemaForm/styles.css
```diff
@@ -0,0 +1,12 @@
+.array-item {
```
stefan-kracht, line 1: What does this file do exactly? Is there any chance we can get rid of it and use tailwind?

stefan-kracht (changes requested):
I'm pretty happy with it Aaron. These (and the comments in code) are the only suggestions that I have

- the headlines are still not the same size/format as our modal headlines
- the descriptions like "Function definition unique identifier." should be a little bit lighter color as the other text
- the + and - buttons in array are swapped in color. The minus should be the destructive one

EDIT: I merged the latest redesign branch into it to solve a conflict, also I will change this from draft to a proper PR

src/design/JSONschemaForm/jsonSchema.ts
```diff
@@ -0,0 +1,232 @@
...
+  description: "A simple form example.",
+  type: "object",
+  required: ["firstName", "lastName"],
+  properties: {
+    firstName: {
+      type: "string",
+      title: "First name",
+      default: "Chuck",
+      contentEncoding: "base64",
+      contentMediaType: "image/png",
```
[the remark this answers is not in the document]
porleaaron: Yes, I think we should delete them

src/design/JSONschemaForm/styles.css
porleaaron: This css defines the size of toolbox which are "remove", "move up", "move down" buttons
The auto-generated code for the form has the Bootstrap classes like **col-xs-9** that actually has no meaning in our code base
So, we should define those classes
We have no way to insert tailwind classes as they are generated by the Form library

stefan-kracht: I wonder if we could still get it to work with tailwind when we have access to some parent.

```html
<div className="[&_.some-class-very-deep-in-the-dom]:border">
  <div>
    <div>
      <div className="some-class-very-deep-in-the-dom">
        I have a border
      </div>
    </div>
  </div>
</div>
``` 

I agree that this is not very intuitive, but I would love to have as little custom CSS files as possible. I also had an issue with playwright the other day when it just could not run when it parsed `import "./somecss.css";` somewhere in the code (which I found super weird because it works in `src/App.tsx`)

src/design/JSONschemaForm/index.stories.tsx
```diff
@@ -1,41 +1,69 @@
...
+export const CustomArraySampleForm = () => (
+  <JSONSchemaForm schema={CustomArraySample} />
+);
+
+export const SampleFormWithFileInput = () => (
+  <JSONSchemaForm schema={SimpleSample} />
+);
+
+export const FunctionSchemaForm = () => (
+  <JSONSchemaForm schema={FunctionSchema as unknown as RJSFSchema} />
```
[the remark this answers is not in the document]
porleaaron: Okay

src/design/JSONschemaForm/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+            <SelectItem key={`select-${op.value}`} value={op.value}>
+              {op.label}
+            </SelectItem>
+          ))}
+        </SelectGroup>
+      </SelectContent>
+    </Select>
+  </div>
+);
+function SubmitButton(props: SubmitButtonProps) {
```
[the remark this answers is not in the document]
porleaaron: yeah

src/design/JSONschemaForm/styles.css
porleaaron: Removed the usage of the style.css
So this is resolved

stefan-kracht (approved):

stefan-kracht (changes requested):
Looks very good Aaron!! I made some minor design changes to align the buttons in the array form and also made the buttons a bit smaller. Since the Submit button has text in it, we would have to introduce useTranslation(), but I would rather not do that in any of your components in the design folder. So I just made sure to never render a submit button at all because we don't have one in our old design as well.

Here are my last tiny tiny tiny change request:

- the description is still in the same color. 
- it would be nice to have some space below "A simple form example description."

Bildschirm­foto 2023-06-09 um 11 00 32

stefan-kracht (approved):

PR #379, opened 2023-05-30
[DIR-581] API Translations

porleaaron:


stefan-kracht (changes requested):
Thank you, Aaron. Just fixed some small typos/copy & paste issues.

In the [ticket](https://linear.app/direktiv/issue/DIR-581/i18n-in-api-folder), I made a snippet with 

` "error": "An error occurred" // you can copy this and use the same for the generic title`

Can you update the code to use `api.generic.error` everywhere where "An error occurred" is used? This would make the translation file much cleaner.

You can also add a `api.generic.undefinedNamespace` for all "namespace is undefined".

stefan-kracht (approved):

stefan-kracht:
Looks good @porleaaron!!

I just noticed that the error "namespace is undefined" was not translated in every api file. I decided to not translate it. Because this is an unexpected error that is not supposed to be shown to the user.

PR #381, opened 2023-05-31
Stefankracht/dir 592 add header support to api factory

stefan-kracht:


sebxian (changes requested):
Thank you, this looks good, but I think a renamed term still needs to be updated in the comments, please check my suggestion. Otherwise approved.

PR #382, opened 2023-05-31
Stefankracht/dir 491 add yaml if user creates a workflow

stefan-kracht:


stefan-kracht:
@sebxian review is not urgent, it's just a ticket that I stumble upon very often and now I quickly wanted to fix it. 

When you create a workflow with a certain `name`, it will now automatically change it to `name.yaml` instead of showing an error toast message.  Because recently, our API has changed so that the file extension now must be provided when creating a workflow. Prior to that, workflows did not have any file extensions at all.

PR #383, opened 2023-06-01
draft: [DIR-198] Settings Page 

sebxian:
This is a draft, but it would be good to get a review at this stage, discuss some decisions and next steps together.


public/locales/en/translation.json
```diff
@@ -164,6 +224,19 @@
...
+        "cancel": "Cancel",
+        "create": "Create",
+        "delete": "Delete",
+        "save": "Save"
+      }
+    },
+    "dialog": {
+      "header": {
+        "confirm": "Confirmation required"
+      }
```
sebxian, line 239: Certain translation strings are re-used in different contexts. I think it would be good to have a "global" section in this file to define them. For now, I put them under "components", although they are not really strictly restricted to these components.

src/api/registries/mutate/createRegistry.ts
```diff
@@ -0,0 +1,61 @@
...
+      urlParams: {
+        namespace: namespace,
+      },
+      headers: undefined,
+    });
+
+  return useMutation({
+    mutationFn,
+    onSuccess: (registry, variables) => {
+      queryClient.invalidateQueries(
```
sebxian, line 48: I postponed implementing an optimistic cache update for later because I wanted to focus on getting this to work first.

Now that it works, I somehow prefer keeping it like this. The code is cleaner, less complex, and refetching the data shouldn't have a large performance impact?

src/api/variables/mutate/updateVariable.ts
```diff
@@ -0,0 +1,70 @@
...
+} from "../schema";
+import { useMutation, useQueryClient } from "@tanstack/react-query";
+
+import { apiFactory } from "~/api/utils";
+import { useApiKey } from "~/util/store/apiKey";
+import { useNamespace } from "~/util/store/namespace";
+import { useToast } from "~/design/Toast";
+import { varKeys } from "..";
+
+const updateVar = apiFactory({
```
sebxian, line 14: This is used for creating and updating variables because the backend API implementation is the same for both cases.

src/api/variables/mutate/updateVariable.ts
```diff
@@ -0,0 +1,70 @@
...
+      },
+      headers: {
+        "content-type": mimeType,
+      },
+    });
+
+  return useMutation({
+    mutationFn,
+    onSuccess: (data, variables) => {
+      queryClient.invalidateQueries(
```
sebxian, line 51: Since this is used for creating and updating, we need to update the cache for the list as well as the single variable. Technically, only one operation is needed in each case, but for now, I don't think it's worth adding that kind of complexity to the code before we update the backend API.

public/locales/en/translation.json
stefan-kracht: Yeah, I thought about that last week too. Maybe we find something on the root level. Maybe call it `common`? We don't have to decide it now, and could make a ticket for aaron to refactor it after this was merged.

src/api/__tests__/apiFactory.test.ts
```diff
@@ -53,7 +53,7 @@ const testApi = setupServer(
   ),
   rest.get(apiEndpointEmptyResponse, (_req, res, ctx) => res(ctx.status(204))),
   rest.get(apiEndpointTextResponse, (_req, res, ctx) =>
-    res(ctx.text("this is a text response"))
+    res(ctx.json({ body: "this is a text response" }))
```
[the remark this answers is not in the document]
stefan-kracht: Yes another two tests would be good. Showing the headers in a test response and also in a json reponse as well.

src/api/utils.ts
```diff
@@ -79,10 +79,15 @@ export const apiFactory =
...
         // We use the text response if its not an empt string
-        if (textResult !== "") parsedResponse = textResult;
+        if (textResult !== "") parsedResponse = { body: textResult };
       }
       try {
-        return schema.parse(parsedResponse);
+        // if response has content, add headers, otherwise return null
+        const headers = Object.fromEntries(res.headers);
+        if (parsedResponse) {
+          return schema.parse({ ...parsedResponse, headers });
```
stefan-kracht, line 88: I wonder if we should not merge the headers into the JSON response and rather divide it into body and headers here as well. 
pros:
- a JSON response would be consistent with the text response
- a JSON response can not have its own `headers` property atm, because it would be overwritten by the headers of the response

cons: 
- we would have to update every response schema to have body and headers

src/api/registries/mutate/createRegistry.ts
stefan-kracht: I agree that it is much cleaner and it's good (at least for now) to keep it simple and just refetch. 

Performance shouldn't be the reason for optimistic updates, I think it's more about skipping a loading state and make the UI a bit more instant. We might even decide to just opt in to optimistic updates when it's about a feature where the user interacts a lot.

src/api/variables/mutate/updateVariable.ts
stefan-kracht: I agree! Maybe this would be worth a comment just to make sure we don't accidentally identify this as a bug in the future

stefan-kracht:
- toast messages should also be in the translation file (you might have copy and pasted that from my code, but Aaron recently refactored that for me, you might need to update merge from `feature/design` first to reuse some of the translations)
- Edit variables: primary button is red
- The bg  color of the editor and the modal don't match, maybe it would kook nicer if we put the editor in a card component like on the workflow page
- When there are no entries in any of the tables, it could look like an empty directory in our explorer (but maybe without the buttons, it could look a bit too crowded?)
- The right side of the table looks very unsettled atm, because of the amount of icons and a missing optical axis
Bildschirm­foto 2023-06-05 um 10 53 08
  - Maybe the icons on the top of the table could have either
    - just an outline
    - outline and text like "Create Secret"
  - Should we opt in to always have a context menu (three buttons icon)?
    - I think I would rather have a context menu with only one entry sometime but therefore have a consistent look throughout the app.

public/locales/en/translation.json
sebxian: `common` sounds good. I'm leaving it as it is right now, but created a ticket ([DIR-601]). I'll assign it to myself for now to keep track.

src/api/__tests__/apiFactory.test.ts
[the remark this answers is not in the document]
sebxian: Added the tests in 0e506ce7ff06b99a986808ab3541bbca66612e3d

src/api/variables/mutate/updateVariable.ts
sebxian: solved in 0e506ce7ff06b99a986808ab3541bbca66612e3d

src/api/utils.ts
sebxian: solved in https://github.com/direktiv/direktiv-ui/commit/435e4b2f04f15fbfb612773753b5c1131e27b94b

sebxian:
To do based on review:

* [x] toasts should use translation strings
* [x] edit variables: primary button color
* [x] card component around editor
* [x] empty list look: Took inspiration from empty folder in explorer, but kept it very minimal (no icons or buttons).
* [ ] right side icons unsettled
* [ ] always use a context menu?

PR #385, opened 2023-06-06
Stefankracht/dir 595 view old revisions

stefan-kracht:


stefan-kracht:
Just updated the back button from a real button to a link that looks like a button (this button feature was just merged into the redesign branch, so I quickly implemented it here)

src/pages/namespace/Explorer/Workflow/Revisions/Overview/TrafficShaping/RevisionSelector.tsx
```diff
@@ -7,16 +7,16 @@ import {
   CommandItem,
   CommandList,
   CommandStaticItem,
-} from "../../../../../../design/Command";
+} from "../../../../../../../design/Command";
```
sebxian, line 10: should we use a `~/` path here instead? (also applies to the other imports in this file)

src/pages/namespace/Explorer/Workflow/Revisions/Overview/TrafficShaping/index.tsx
```diff
@@ -1,14 +1,14 @@
 import { FC, useEffect, useState } from "react";
 import { Trans, useTranslation } from "react-i18next";
 
-import Button from "../../../../../../design/Button";
-import { Card } from "../../../../../../design/Card";
+import Button from "../../../../../../../design/Button";
```
sebxian, line 4: should we use a `~/` path here instead? (also applies to the other imports in this file)

sebxian (commented):
## Functional review

One detail that's a bit confusing:

When I'm in the "Active Revision" tab and make a revision, then click on the "Open Revision" in the toast, I'm still in the "Active Revision" tab but the editor is read-only. I was surprised by that, intuitively the "Active Revision" tab should always be editable and I should be on the "Revisions" tab when viewing an old revision.

Also in general, I think it would be good to have a more obvious indication when the editor is "read only".

Code review: No issues, except a few imports could be updated to use the `~/` pattern.

I haven't looked at the specs file in detail. I still have lots of errors when running the e2e tests on my machine. I'll investigate this and leave another comment in case I find issues with that.

stefan-kracht:
> When I'm in the "Active Revision" tab and make a revision, then click on the "Open Revision" in the toast, I'm still in the "Active Revision" tab but the editor is read-only. I was surprised by that, intuitively the "Active Revision" tab should always be editable and I should be on the "Revisions" tab when viewing an old revision.

This is a "separate" issue that I track [here](https://linear.app/direktiv/issue/DIR-551/update-active-tab-in-workflow). Basically the tabs never update their active state when you navigate to one of them without explicitly clicking on them (f.e. when you open a revision with the button on a toast notification). I will solve that separately.

> Also in general, I think it would be good to have a more obvious indication when the editor is "read only".

I removed the native readonly hint because it looks ugly but I did not added a cusom one. But now I did (see second screenshot below)

**Old**

Bildschirm­foto 2023-06-09 um 17 15 18

**Just added this**

Bildschirm­foto 2023-06-09 um 17 13 38

I also [just made the Button labels shorter](https://github.com/direktiv/direktiv-ui/pull/385/commits/b6f691cbeb5e046697518370d9d73639f61bf88b) because it did not work well on smaller screens. The revision hashes are so long that they linebreak very early.

sebxian (commented):
I made some small changes in ca76073d0fe0f0ac6851e3f66eb617fa7a0fa59e.

Not a show stopper, but the tooltip in the read-only editor is a bit annoying, because its visibility is toggled by a mouse click. That means, every time I click in the editor (maybe to select text for copying or so) it appears or vanishes. Or maybe this is good because the tooltip would be easy to miss otherwise?

Otherwise approved.

stefan-kracht:
> Not a show stopper, but the tooltip in the read-only editor is a bit annoying, because its visibility is toggled by a mouse click. That means, every time I click in the editor (maybe to select text for copying or so) it appears or vanishes. Or maybe this is good because the tooltip would be easy to miss otherwise?

Thanks! I agree, we could definitely evaluate this behavior in the future but keep it for now.

I [just added](https://github.com/direktiv/direktiv-ui/pull/385/commits/92fb43d7b6b287985c25fa189236fc573f7c0d98) a class to the tooltip to avoid linebreaks

PR #386, opened 2023-06-06
[DIR-597] extend button component to work with any tag

porleaaron:


stefan-kracht:
This looks good. Because `asChild` will not work together with the `loading` prop, I added some Typescript that will reflect that. So now, we can not accidentally use them together.

PR #387, opened 2023-06-07
[DIR-461] create workflow dialog ui update

porleaaron:


stefan-kracht:
made some mini layout changes, but other than that 🥳

PR #389, opened 2023-06-09
[DIR-594] Second Approach ButtonBar component

porleaaron:


stefan-kracht:
I like this one much better. It's much simpler, and we can use it even more flexible

- Can we remove the `[&_button]:active:scale-100` part? It feels weird when all of our buttons scale a bit when you click on them, but they don't. 
- Is it possible to make the border in the middle only 1px? We have to solve it in a way that the border is still there but overlap (so that when a button scales when you click on it, will always have a border on both sides)
- I think we should have a couple of more stories to showcase how flexible this new wrapper is. Some that came into my mind:
  - making a "toolbar" with more than 4 or 5 buttons next to each other
  - showcase all three sizes
- The last step would be to integrate the new button on the workflow page as you did in our first approach

stefan-kracht:
This is what I meant with:

> Is it possible to make the border in the middle only 1px? We have to solve it in a way that the border is still there but overlap (so that when a button scales when you click on it, will always have a border on both sides)

Bildschirm­foto 2023-06-13 um 10 18 56
Bildschirm­foto 2023-06-13 um 10 19 01

- I fixed it [here](https://github.com/direktiv/direktiv-ui/pull/389/commits/90aceb1a37733701e0a492d7db533f83e6580e84)
- I also removed the Toolbar story because I found it misleading. Our stories should show suggestions on how to use the components and I think we should not encourage to mix button sizes and variants inside a toolbar

PR #390, opened 2023-06-13
[DIR-614] Card Component background update

porleaaron:


stefan-kracht:
Good work Aaron. Sebastian and I just discussed the Card API a little bit and we slightly changed it to be semantically more explicit. We agreed on make the changing the `weight = 1 | 2 | 3` to `background = "none" | "weight-1" | "weight-2"`. I also reviewed every Card usage and remove the bg when possible.

PR #393, opened 2023-06-13
[DIR-608] add syntax highlighting styles for more languages

porleaaron:


stefan-kracht:
I think this looks VERY good already!

I just have one thing that bugs me a bit. When comparing the editor in light and dark mode, at some places they seem to have a different behavior. It would be nice if both would have the same "pattern" (even if the colors are not the same).

It's easier to explain it with an example. In the html editor
- A color like `#a0a0a0` in the inline CSS should also have the color of a comment like in the light theme
- A metric like `40px` should also have this primary (blue) color like in the light them
- there are similar differences in the CSS editor as well

These are the only issues that I have. It would be nice if you can look into them.

PS: I split the themes into two separate files. I think it makes it easier to compare both to each other.

stefan-kracht:
This looks super nice now!

PR #394, opened 2023-06-15
[DIR-616] basics of fileInput component

porleaaron:
I need your review
For this component, we have a wrapper, two buttons(left and right section of the input), a hidden input
I introduced the combinedRef which should combine the Refs

I would like to have a feedback on this structure


stefan-kracht:
One issue I have is, that we should not have text in our design components that need to be translated. So in this current example, we would have to make the text "choose file" be passed by a prop.

Wouldn't it be possible to just style the input and not introduce any button? Like in [this example](https://flowbite.com/docs/forms/file-input/).

porleaaron:
@stefan-kracht I removed the new component

Just modified the original Input component to handle "type" attribute for "file"

stefan-kracht:
Good job AAron. I was so fixated that this had to be a separate component, that I did not think about it being part of our Input-Component. This makes much more sense. I just made one [tiny tiny tiny change](https://github.com/direktiv/direktiv-ui/pull/394/commits/b9c7e6037d6973679f5f1a2c1eab153d357a0f81) to make the styles a bit more structures/readable.

PR #395, opened 2023-06-15
[DIR-617] Logs, LogEntry component

porleaaron:
Should we copy monaco font or should we use our theme fonts in the Monaco editor?


stefan-kracht:
Thanks Aaron!

-  can you remove the white space in the wrap examples? I thought this was a bug. Just make some long lines without whitespace.
- I updated "Editor VS Logs Font Compare" story to be more realistic, because you just used one line instead of multiple ones. Now you can see that we need to adjust the padding a bit to get tot the same line height. The Card of the editor has a padding of 4 because it looks better with some space, however the Card of the logs has no padding because it would look weird when the log entries don't touch the edge. So I think it is okay that the padding is a bit different here. 
- I added a button to the "Editor VS Logs Font Compare", can you check why the logs take up so much space when disabling line wrap?

PR #397, opened 2023-06-15
[DIR-619] Settings e2e tests

sebxian:


e2e/settings/index.spec.ts
```diff
@@ -0,0 +1,208 @@
...
+  await page.getByTestId("new-registry-pwd").type(newRegistry.password);
+  await page.getByTestId("new-registry-user").type(newRegistry.user);
+
+  await page.getByTestId("registry-create-submit").click();
+  await actionWaitForSuccessToast(page);
+
+  const menuButtons = page.getByTestId(/dropdown-trg-item-/);
+  await expect(menuButtons, "number of menuButtons should be 4").toHaveCount(4);
+  const itemName = page.getByTestId("item-name");
+  const removing = await itemName.nth(2).innerText();
```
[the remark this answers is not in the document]
sebxian: @porleaaron I already updated this in the code, so just FYI.

e2e/settings/index.spec.ts
```diff
@@ -64,7 +64,7 @@ test("it is possible to create and delete secrets", async ({ page }) => {
   await page.goto(`/${namespace}/settings`);
   await page.getByTestId("secret-create").click();
   const newSecret = {
-    name: faker.random.alphaNumeric(7),
+    name: faker.internet.domainWord(),
```
sebxian, line 67: @porleaaron I had to change this because the name needs to start with a letter. When the first letter is a number, it will fail to create the secret in the backend.

It also looks a bit nicer, in case we want to share screenshots or screencasts from our tests with other people.

e2e/settings/index.spec.ts
```diff
@@ -0,0 +1,262 @@
...
+    "initial textarea value should be the variable's content"
+  ).toBe(subject.content);
+  await expect(
+    page.locator("select"),
+    "MimeTypeSelect is set to the subject's mimeType"
+  ).toHaveValue(subject.mimeType);
+
+  // This was needed previously to make sure the editor is initialized
+  // before updating the value, but it should no longer be needed.
+  // await page.getByTestId("variable-editor-card").click();
```
sebxian, line 236:
```suggestion
```

@porleaaron did you check if this line is still needed? If it works fine without the .click() action, we should remove these three lines.

sebxian (commented):
@porleaaron thanks again for the good work writing these tests :+1: 

As we discussed in Linear, there were a few timing problems with the Editor introduced by changes we have made in other branches. It took some time to figure these out but they should be fixed now.

Other than this, there are a few comments that I would ask you to check. Most of them are related to the same problem.

If you agree with my suggested changes, feel free to use the GitHub UI to apply them. If you have any comments or questions, let me know.

PR #398, opened 2023-06-19
[DIR-628] Settings: Fixes in create/edit variable forms

sebxian:
This addresses some comments from the review in [DIR-198]

Changes:
* Moves the mimetype select above the editor for creating/editing variables
* Fix binding between mimetype and editor language when creating variables (already worked when editing)
* Set default mimetype to json (like it was in the old ui)
* Fix bindings between labels and inputs in forms

I also thought about consolidating the `Create` and `Edit` components because they are very similar, but there are also many differences to keep track of, most of which involve hooks. So it gets really complicated and I think it's better to leave them as is for now.


src/pages/namespace/Settings/Variables/Create.tsx
```diff
@@ -5,47 +5,58 @@ import {
...
       content: body ?? "",
-      mimeType: mimeType ?? "text/plain",
+      mimeType: mimeType ?? "application/json",
     },
   });
 
+  useEffect(() => {
+    setEditorLanguage(mimeTypeToLanguageDict[mimeType]);
+  }, [mimeType]);
+
```
stefan-kracht, line 59: I would suggest removing this `useEffect` here and introduce something like a `onMimetypeChange` function that takes care of changing both variables.

```typescript
const onMimetypeChange = (value: MimeTypeType) => {
    setMimeType(value);
    setEditorLanguage(mimeTypeToLanguageDict[value]);
};
``` 
and wire this function to the MimeTypeSelect component.
```typescript
<MimeTypeSelect
  id="mimetype"
  mimeType={mimeType}
  onChange={onMimetypeChange}
/>
``` 

### Why?
In my understanding, `useEffect` should only be used to react to a change that is not controlled by the component explicitly or happens asynchronously (like waiting for data to be fetched), but in this case the effect reacts to a `mimeType` change which is fully controlled by this component.

stefan-kracht (changes requested):
Just have two tiny tiny things.

src/pages/namespace/Settings/Variables/MimeTypeSelect.tsx
```diff
@@ -38,14 +38,16 @@ export const MimeTypeSchema = z.enum([
...
   onChange,
 }: {
+  id?: string;
   mimeType: string | undefined;
   onChange: (value: MimeTypeType) => void;
 }) => (
   <Select value={mimeType} onValueChange={onChange}>
-    <SelectTrigger variant="outline">
+    <SelectTrigger id={id} variant="outline" block>
       <SelectValue placeholder="Select a mimetype" />
```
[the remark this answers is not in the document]
sebxian: Good catch, updated in new commit.

src/pages/namespace/Settings/Variables/Create.tsx
sebxian: I updated this as suggested and also made similar changes in the Edit component.

While doing this, I also updated the defaults for mimeType / language. 

They are now intentionally different in Create and Edit. In Create, for new variables, the default is json, like in the old UI. 

In Edit, I renamed the variable to fallbackMimeType and set it to text/plain. This takes effect when an existing variable has a mime type that is not implemented, like for example, XML. This can happen when you upload a file (in the old UI, then the extension determines the mime type).

In DIR-624, we might find a better solution for how to deal with file uploads that result in unsupported mimeTypes.

stefan-kracht (approved):

PR #399, opened 2023-06-20
[DIR-618] rename apiFactory file, make headers and payload optional

sebxian:
Changes:
* src/api/utils.ts renamed to src/api/apiFactory.ts to make it easier to find
* apiFactory params headers and payload are now optional. In an earlier state of the code, it was necessary to declare them as `undefined` for reasons related to type inference. This has changed in the meantime. This PR makes them optional and removes unnecessary declarations.


stefan-kracht (approved):
Looks good. I just pushed [one variable name change](https://github.com/direktiv/direktiv-ui/pull/399/commits/f034523f66efea344142f569739add0a50ce3315) that bugged my for a while now.
Feel free to merge, when you checked my change.

PR #400, opened 2023-06-20
[DIR-627] Settings: loading states

sebxian:
Changes:
* Lists on settings page are not rendered before they are fetched to avoid spinners / layout shifts 
* When editing a variable, the editor is not rendered before the content is fetched, and the mimeTypeSelect has a loading spinner.


stefan-kracht (approved):
This looks 👌

PR #401, opened 2023-06-21
[DIR-629] top right navigation components

porleaaron:


stefan-kracht (changes requested):
Hi Aaron. Here is some feedback on how we can improve the component an e2e tests a bit.

- the `hasAccount` and `username` should not be a prop. It can be a const at the very top of the file outside the component. Later we will replace it with a hook.
- all text must move into translation files
- on all `_blank` links, it is good to use `rel="noopener noreferrer"`
- we need a border above "Appearance" when `isAdmin` is `true`
- `twMerge` should be called where `clsx` is used

The e2e tests could be a bit more specific and also test what exactly is happening so that it breaks if the dark mode is not working. 

You could do it like:
- at the beginning the text in the menu contains "light mode"
- the background color of the body is X
- then you click on it
- now the text in the menu contains "dark mode"
- the background color changed to Y
- then you click on it again
- now the text in the menu contains "light mode" again
- and the background color changed to X again

Generally, the second parameter of `expect` should be very specific as well. F.e. "this mode has the pattern" should be "the menu item contains the word 'light mode'".

porleaaron:
This PR is updated according to your last comment @stefan-kracht

stefan-kracht:
Here are some changes that I made
- The word "Help" was not translated
- swapped [NavigationLink](https://feature-redesign--6418690f3c0c39f7a2975666.chromatic.com/iframe.html?args=&id=components-navigationlink--navigation) against a simple a tag (NavigationLink should only be use in the main navigation for now, or if we would use it there, than all links should be a NavigationLink to avoid overwriting the padding)
- I used the chance to also update the `NavigationLink` component to have a much simpler interface
- I [simplified the e2e test](https://github.com/direktiv/direktiv-ui/pull/401/commits/3a068805a5e0dc69ce6f374c31bdceecfa21755c) a bit and made them a bit more explicit

stefan-kracht (approved):

PR #405, opened 2023-06-23
[DIR-630] settings editor focus

sebxian:
After DIR-542 has been merged, the Editor components used for creating secrets and variables no longer focus on mount. This is as intended, because the name input should be focused first.

When editing variables, it makes sense to focus the Editor right away. This PR adds that behavior.

Preview at https://direktiv-ui-refs-pull-405-merge.direktiv.dev/sebxian/settings


stefan-kracht (approved):
Looks great!! That was easy to review with the preview link. Especially because I was also able to use a preview link of the current version to compare the behavior.

PR #407, opened 2023-06-23
Draft: [DIR-632] New Instances Page (mvp)

stefan-kracht:
### [🚀 Preview 🚀](https://direktiv-ui-refs-heads-stefankracht-dir-632-instances-page.direktiv.dev/)

- Router was updated to handle the list and detail page
- Breadcrumb now has a different behavior on the instances page
  - I split it up into more components, because it will grow even further over time
- I changed the instances icon in the main menu to reflect instances (plural) instead of one instance. This makes more sense, especially yin the breadcrumb
Bildschirm­foto 2023-06-28 um 11 04 15

Some things are a bit out of scope of the main ticket, but needed to be done to implement the list as it is:
- I decided to simplify all copy and paste buttons to have no text at all. At some places it is just easier to have a very tiny button and I rather like it to be consistent instead of having two variations of that button (I think our main target group will understand the copy and paste button with the icon and not text)
- fixed a book in `useUpdatedAt` hook where the initial render might show a wrong state
- added a `ConditionalWrapper` component (it's self explaining if you look at the usage in `src/pages/namespace/Instances/List/Row.tsx`)
- Design components
  - Breadcrumb need a small update on spacing between ([old](https://direktiv-ui-refs-heads-feature-redesign.direktiv.dev/dd/explorer/tree/oejoiejde/ledmwe) vs [new](https://direktiv-ui-refs-heads-stefankracht-dir-632-instances-page.direktiv.dev/dd/explorer/tree/oejoiejde/ledmwe))
  - added a success variant to our Badge (`<Badge variant="success">success</Badge>`)
  - HoverCard now has a `noBackground` prop to make it possible to place an Alert component inside of it
  - Pagination
    - remove the background
    - reflect the `disabled` state of a pagination button

The following tasks will be done separately:
- [add an abstraction layer to the pagination](https://linear.app/direktiv/issue/DIR-650/instances-pagination)
- [add e2e tests](https://linear.app/direktiv/issue/DIR-651/instances-e2e-test)
- [filter will be added](https://linear.app/direktiv/issue/DIR-652/instances-filter)


stefan-kracht:
> * In the left navigation bar, "instances" is not capitalized. This is not new, it's already so in `feature/redesign`. But now might be a good time to fix it.

I updated the instances to be uppercase. It was not even in the translation files yet! I also made a small ts change. There was an optional property, which is not optional anymore.

> For example, we could use package for registries. 

Which one is that?

Bildschirm­foto 2023-06-28 um 16 26 15

sebxian:
> Which one is that?
> 
> Bildschirm­foto 2023-06-28 um 16 26 15

https://lucide.dev/icons/package

Sorry, I tried to link it in my comment but GitHub ate the markdown syntax.

PR #410, opened 2023-06-27
[DIR-620] settings: broadcasts

sebxian:
This adds broadcasts to the settings page.


public/locales/en/translation.json
```diff
@@ -218,6 +218,25 @@
...
+          "fail": "Failed",
+          "create": "Created",
+          "update": "Updated",
+          "delete": "Deleted"
+        },
+        "rows": {
+          "instance": "Instance",
+          "directory": "Directory",
+          "workflow": "Workflow",
+          "instance-variable": "Instance Variable",
```
stefan-kracht, line 235: I think we want the row keys to be in lowerCamelCase as well?

src/api/broadcasts/query/useBroadcasts.ts
```diff
@@ -0,0 +1,39 @@
...
+    apiKey,
+    urlParams: { namespace },
+  });
+
+export const useBroadcasts = () => {
+  const apiKey = useApiKey();
+  const namespace = useNamespace();
+
+  if (!namespace) {
+    throw new EvalError("namespace is undefined");
```
stefan-kracht, line 29: I think this was a VSCode autocomplete thing, and it should be `Error` instead of `EvalError`?

src/pages/namespace/Settings/Broadcasts/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+        <h3 className="flex items-center gap-x-2 font-bold text-gray-10 dark:text-gray-dark-10">
+          <Radio className="h-5" />
+          {t("pages.settings.broadcasts.title")}
+        </h3>
+      </div>
+
+      <Card>
+        <Table>
+          <TableHead>
+            <TableRow className={rowClasses}>
```
stefan-kracht, line 54: I was wondering if we disable the hover effect on the table header rows? I don't know if it's just me, but I only expect this hover effect to be available on rows that I can interact with.


If you agree, it should work with this code (I just used it myself)

```typescript
import { twMergeClsx } from "~/util/helpers"; // this will work when you update the base branch of your PR
...
<TableRow
  className={twMerge(
    rowClasses,
    "hover:bg-inherit dark:hover:bg-inherit"
  )}
>


```

src/pages/namespace/Settings/Broadcasts/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+      payload: {
+        broadcast: {
+          [key]: !broadcasts[key],
+        },
+      },
+    });
+
+  return (
+    <>
+      <div className="mb-3 flex flex-row justify-between">
```
stefan-kracht, line 44: The headline is a bit too close to the Card compared to the other components on the page. I guess the reason for that is the missing button on the top right.

src/pages/namespace/Settings/index.tsx
```diff
@@ -1,3 +1,4 @@
+import Broadcasts from "./Broadcasts";
 import { FC } from "react";
```
stefan-kracht, line 2: I just had an idea, and I would love to hear your feedback on this (because I'm not sure about it yet). The settings page has a lot of components and each of them load data and will populate its content when all data was fetched. Each of this data population is very likely to happen in its own render loop. So there is a lot of flickering on the initial page load.

What do you think about the approach to wait for all data to be fetched before rendering any component? From the user perspective, it just removes the flickering, but I'm unsure about the DX/maintainability of this code:

```typescript
import Broadcasts from "./Broadcasts";
import { FC } from "react";
import RegistriesList from "./Registries";
import SecretsList from "./Secrets";
import VariablesList from "./Variables";
import { useBroadcasts } from "~/api/broadcasts/query/useBroadcasts";
import { useRegistries } from "~/api/registries/query/get";
import { useSecrets } from "~/api/secrets/query/get";
import { useVars } from "~/api/variables/query/useVariables";

// this hook uses all initial data fetching hooks from the components on
// this page and returns true if all of them are fetched. This is used to
// render the page only when all data is fetched and avoid flickering as
// all components are rendered at the same time and data comes in at different
// times.
const usePrefetchData = () => {
  const { isFetched: isFetchedisFetchedBroadcasts } = useBroadcasts();
  const { isFetched: isFetchedSecrets } = useSecrets();
  const { isFetched: isFetchedRegistries } = useRegistries();
  const { isFetched: isFetchedVars } = useVars();
  return (
    isFetchedisFetchedBroadcasts &&
    isFetchedSecrets &&
    isFetchedRegistries &&
    isFetchedVars
  );
};

const SettingsPage: FC = () => {
  const prefetched = usePrefetchData();

  if (!prefetched) {
    return null;
  }

  return (
    <div className="flex flex-col space-y-6 p-10">
      <section>
        <SecretsList />
      </section>

      <section>
        <RegistriesList />
      </section>

      <section>
        <VariablesList />
      </section>

      <section data-testid="broadcasts-section">
        <Broadcasts />
      </section>
    </div>
  );
};

export default SettingsPage;
``` 

**PRO**
- It is independent of the components and just prefills the cache, each of the components would still work perfectly fine with or without this code.

**CON:**
- this hook needs to be maintained, it's easy to forget about it when updating any component on that page (even if the possible "damage" will not break anything, but could make one or two request too much or too little)

stefan-kracht (changes requested):
Very good, I like this a lot. It has a clean UI and the internal state management is super tiny because it's directly wired up to the backend.

I just added some very minor feedback and a small discussion about data fetching in general.

public/locales/en/translation.json
sebxian: fixed

src/api/broadcasts/query/useBroadcasts.ts
sebxian: Yes, good catch. Fixed.

src/pages/namespace/Settings/Broadcasts/index.tsx
sebxian: Agreed and fixed :-)

src/pages/namespace/Settings/Broadcasts/index.tsx
sebxian: That's correct. I added padding to the headlines so they have exactly the same height as the button in c973c17ffb35429b9e879a32483c3191c0c86763

src/pages/namespace/Settings/index.tsx
sebxian: I think this is a clever idea and an elegant solution. I think the maintainability drawback is very minor because the hook is very easy to understand.

However, I was a bit worried that the page could be empty for too long if we wait until all data is fetched, so I played around with bandwidth throttling in devtools.

But even on slow 3G, all data is rendered almost instantly (it's just a few Kb that need to be fetched). But that also means there isn't a lot of noticeable flickering because everything loads so quickly. (The initial loading of the assets, however, takes very long, but that is a different story.)

In conclusion, I have no strong objections, but I am a bit hesitant about this because it adds more complexity to solve a problem that may not really exist.

stefan-kracht: Thanks for that feedback. Personally, the flickering bothers me enough to actually vote for that caching solution.

sebxian: Ok, implemented in 24b00dbb465acbdcc314e9b82c82736bb6079e33.

stefan-kracht (approved):

stefan-kracht:
Feel free to merge at any time 👌

PR #412, opened 2023-06-29
[DIR-650] instances pagination

stefan-kracht:
This PR implements the pagination on the instances list page

- It comes with a new Pagination component that can be used for other pagination as well
- I changed the instances list to show 15 per default instead of just 10 (took my 14-inch MacBook as reference for screen space)
- remove hover on table row head

Unfortunately, no preview link could be generated, because of the branch name length

_if you want to quickly generate a lot of instances, just open a workflow in the UI. And when you are running it, find the POST request in the dev tools, right-click on the request and select copy / copy as cURL. Now run this line in the terminal and spam arrow up and enter to rerun this a couple of times._


src/componentsNext/Pagination/__tests__/describePagination.test.ts
```diff
@@ -0,0 +1,137 @@
...
+    });
+
+    test("page 2 / 10", () => {
+      const pagination = describePagination({ pages: 10, currentPage: 2 });
+      expect(pagination).toStrictEqual([1, 2, 3, "...", 9, 10]);
+    });
+
+    test("page 5 / 10", () => {
+      const pagination = describePagination({ pages: 10, currentPage: 5 });
+      expect(pagination).toStrictEqual([1, 2, "...", 4, 5, 6, "...", 9, 10]);
```
sebxian, line 19: This is ok in my opinion, but since we talked about it earlier, it would also make sense to render the 3 in place of the ellipsis.

src/componentsNext/Pagination/__tests__/describePagination.test.ts
```diff
@@ -0,0 +1,137 @@
...
+    });
+
+    test("page 6 / 10", () => {
+      const pagination = describePagination({ pages: 10, currentPage: 6 });
+      expect(pagination).toStrictEqual([1, 2, "...", 5, 6, 7, "...", 9, 10]);
```
sebxian, line 24: ... and in this case, one could render the 8.

So it seems you need 11+ pages to get both gaps.

src/componentsNext/Pagination/__tests__/describePagination.test.ts
```diff
@@ -0,0 +1,137 @@
...
+
+      const oneNeighbourPagination = describePagination({
+        pages: 10,
+        currentPage: 1,
+        neighbours: 1,
+      });
+      expect(defaultPagination).toStrictEqual(oneNeighbourPagination);
+    });
+
+    test("increasing neighbours will eventually get rid of the ...", () => {
```
sebxian, line 62: A very minor wording suggestion: Should we use "ellipsis" to describe the dots?

We could also use the ellipsis character `…` everywhere, instead of three dots in the code. Technically this is more correct, but I don't think it matters a lot. Also, the ellipsis character is a bit wider when rendered in the DOM, so we would need to update the styling as well.

src/componentsNext/Pagination/describePagination.ts
```diff
@@ -0,0 +1,143 @@
...
+   * endSegmentRight = 10
+   * endSegmentCount = 2
+   * endSegment = ["...", 9, 10]
+   *
+   * and this this function will return
+   * [1, 2, "...", 5, 6, 7, "...", 9, 10]
+   */
+
+  // active segment
+  const currentLeft = current - neighbours;
```
sebxian, line 73: I would suggest `leftmostNeighbour` here (and `rightmostNeighbour` below, respectively)?

(don't forget to update this in the comments as well, if you change it)

sebxian (changes requested):
Very nice work :-)

As we talked about in person, there was a bug when updating the DOM due to a react element key not being unique. This was fixed in 97ce90fb98345f09cd5e6104c59866b570659f9f. 

As we also talked about, in some cases the ellipsis just replaces one single digit, so that digit could be rendered instead. But I would be ok with leaving it as is, since I believe it looks better (even if it is not perfectly logical).

Other than that, see the comments for a few wording suggestions.

src/componentsNext/Pagination/describePagination.ts
stefan-kracht: You mean `currentLeft` should be `leftmostNeighbour`, right? I'm just asking because there are no comments that I can update...

src/componentsNext/Pagination/__tests__/describePagination.test.ts
stefan-kracht: I think this is a good idea and I actually would change it to the `…` sign as you suggested (and I would not worry about the spacing)

src/componentsNext/Pagination/describePagination.ts
stefan-kracht: Changed it as we discussed in person https://github.com/direktiv/direktiv-ui/pull/412/commits/02902336c694f2ebc66663d3fe4e1af8e27ea59e

src/componentsNext/Pagination/__tests__/describePagination.test.ts
stefan-kracht: I really like to not render the ellipsis when it only hides one page. https://github.com/direktiv/direktiv-ui/pull/412/commits/1315ad16d481d7e7f03a7baa2b5ec6b099feaa99

src/componentsNext/Pagination/__tests__/describePagination.test.ts
stefan-kracht: I updated the tests to test everything against 11 pages to cover the use case of two ellipsis https://github.com/direktiv/direktiv-ui/pull/412/commits/e5567f8575cbef6a920f18669e0c17ba6ce66a95

src/componentsNext/Pagination/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+        onClick={() =>
+          previousPage && setOffset(setOffsetByPageNumber(previousPage))
+        }
+        disabled={isFirstPage}
+      />
+      {paginationDescription.map((page, index) => {
+        if (page === "…")
+          return (
+            <PaginationLink key={index} disabled>
+              ...
```
sebxian, line 49:
```suggestion
              …
```

src/componentsNext/Pagination/__tests__/describePagination.test.ts
```diff
@@ -0,0 +1,141 @@
...
+
+      const result3Neighbours = describePagination({
+        pages: 10,
+        currentPage: 3,
+        neighbours: 3,
+      });
+      expect(result3Neighbours).toStrictEqual([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]);
+    });
+
+    test("when neighbours is very high, it will just not having any ellipsis", () => {
```
sebxian, line 89:
```suggestion
    test("when neighbours is very high, it will not have any ellipsis", () => {
```

src/componentsNext/Pagination/index.tsx
stefan-kracht: Damn. I thought this was a change that was 100 % covered by typescript an i just had to update all the places where TS was screaming at me. But I was wrong. Should we use {page} instead?

```typescript
 if (page === "…")
   return (
     <PaginationLink key={index} disabled>
       {page}
```

sebxian: Yes, it makes a lot of sense to use `{page}`. 

Actually, that makes me wonder if it would be better to not to have an if-block for the ellipsis at all. This works (but note the added complexity in the onClick statement necessary to avoid TS errors)

```tsx
// remove the if-block for if (page === "…"), then below that:
        const isActive = currentPage === page;
        const isEllipsis = page === "…";
        return (
          <PaginationLink
            key={index}
            active={isActive}
            onClick={() => {
              !isEllipsis &&
                !isActive &&
                setOffset(setOffsetByPageNumber(page));
            }}
            disabled={isEllipsis}
          >
            {page}
          </PaginationLink>

stefan-kracht: Good idea, that is much cleaner!

PR #416, opened 2023-07-05
[DIR-654] broadcasts e2e test

porleaaron:


e2e/settings/index.spec.ts
```diff
@@ -275,39 +275,36 @@ test("it is possible to edit variables", async ({ page }) => {
   ).toHaveValue(updatedType);
 });
 
-test("it is possible to handle broadcasts", async ({ page }) => {
-  const { broadcast } = await createBroadcasts(namespace);
+test("it is possible to update broadcasts", async ({ page }) => {
+  const { broadcast: broadcasts } = await createBroadcasts(namespace);
```
sebxian, line 279: I renamed `broadcast` to `broadcasts` in this context because it is an array, so it makes more sense to have a plural noun for it. Unfortunately, it is singular in the API response, but at least in this test this will improve readability.

e2e/settings/index.spec.ts
```diff
@@ -275,39 +275,36 @@ test("it is possible to edit variables", async ({ page }) => {
...
+  expect(broadcasts, "test data has been created").toBeTruthy();
```
sebxian, line 280: I moved this up so the steps from setting up the test to visiting the page are logically separated.

Did I understand correctly that you wanted to check setting up the test data was successful?

I updated the text accordingly. The text `broadcast should be truthy` just repeats what is already in the code.

If we add additional test descriptions, we should try to explain our intention in more general, less technical terms. For example, explain why we are doing this, rather than what we are doing.

e2e/settings/index.spec.ts
```diff
@@ -275,39 +275,36 @@ test("it is possible to edit variables", async ({ page }) => {
...
+  // check the initial state
   for (let i = 0; i < BroadcastsSchemaKeys.length; i++) {
     const key = BroadcastsSchemaKeys[i] || "directory.create";
     const checkbox = page.getByTestId(`check.${key}`);
     const isChecked = await checkbox.isChecked();
-    expect(
-      isChecked,
-      `${key}-checkbox status should be the same as in the broadcast`
-    ).toBe(broadcast[key]);
+    expect(isChecked, `checkbox for ${key} has the expected value`).toBe(
```
sebxian, line 289: Again, I edited the description to be a bit more general.

"the same as in the broadcast" is only clear if you read the code and know how the variables are named. 

The test log should be easy to understand without knowing the code.

e2e/settings/index.spec.ts
```diff
@@ -275,39 +275,36 @@ test("it is possible to edit variables", async ({ page }) => {
...
 
-  for (let i = 0; i < randomElements.length; i++) {
-    const key = randomElements[i] || "directory.create";
-    const checkbox = page.getByTestId(`check.${key}`);
-    const isChecked = await checkbox.isChecked();
-    expect(
-      isChecked,
-      `${key}: updated checkbox status should be the invert of the one in the broadcast`
-    ).toBe(!broadcast[key]);
+    // expect() without .poll() would fail, because it would not wait for the DOM to update
```
sebxian, line 302: I updated this so it will first click, then check that the DOM has been updated, and then the next click. This way, everything can be done in one for loop.

I also found out how to make the expect() method wait until the value is the expected value. This was quite tricky.

But it means we don't need to wait manually after clicking, which makes the test a bit quicker and better to read.

sebxian (commented):
Very good work, @porleaaron !

You picked a good approach to randomize the data and the tests look very solid.

At first, I thought we could use `array.forEach()` or `Promise.all(array.map())` instead of for loops. But after trying this out it leads to all kinds of problems with the timing of the tests.

So the for loops seem to be a very good choice in async code.

Since I needed to figure a few things out for myself I already made some changes to the code. You can check them out in 965611339d97cec47e3edaf76f82499388a6ddf2 and read my comments.

I could manage to improve the end of the test a bit. Other than that, I made some changes for better readability.

Please look at these changes and let me know what you think.

porleaaron:
I think your updates are all good

stefan-kracht:
@sebxian I [made a suggestion](https://github.com/direktiv/direktiv-ui/pull/416/commits/49a288219fc5454097f9c76cdc721161ebdc9e97) on how we can dynamically generate a schema. Thanks to TypeScript it was not so straight forward as it should be :) Feel free to update variable names or even revert the commit completely.

@porleaaron I made a tiny change in your test to replace your `pickRandomElements` function against the `faker.helpers.arrayElements`. I appreciate your implementation, but I bet you just did not know that faker has these on board (with [a lot of other cool helpers](https://fakerjs.dev/api/helpers.html) btw!). For a cleaner code base, I think it's best to just use the faker ones as we already have the package anyway

sebxian:
@stefan-kracht thanks for the changes, that looks very good :rocket:

PR #417, opened 2023-07-05
[DIR-653] Maintop to be sticky

porleaaron:
Preview at https://direktiv-ui-refs-heads-porleaaron28-dir-653-sticky-breadcrumb.direktiv.dev/

### DIFF
--- a/src/design/Appshell/index.stories.tsx
+++ b/src/design/Appshell/index.stories.tsx
@@ -364,7 +364,7 @@ export const MoreDetailedShell = () => {
             </MainTopRight>
           </MainTop>
           <MainContent>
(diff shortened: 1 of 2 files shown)


stefan-kracht:
Great work Aaron!

I just slightly changed the AppShell Story. You made the top bar the size of the whole page to see the sticky header, but I think we should instead fill it with content to simulate the behavior.

Otherwise, the background is all over the page
Bildschirm­foto 2023-07-06 um 14 52 43

Bildschirm­foto 2023-07-06 um 15 30 14

I think you missed this task
>  On smaller screens the breadcrumb should not be sticky

but I quickly just added a `lg:` to your new added classes 

Otherwise I will merge it directly 👌

stefan-kracht (approved):

PR #419, opened 2023-07-17
[DIR-652] Instances Filter

sebxian:
Preview link: https://direktiv-ui-refs-heads-sebastianarmbrust-dir-652-inst-fltr.direktiv.dev/sebxian/instances

## Description

* Adds filters to instances. Where the old UI had input fields that were always visible, the new UI only has an "Add filter" button that opens a menu. In the menu, users can in two steps (1) select the field to be filtered and (2) enter a value.
* Filters that have been set appear as buttons above the instances table. The value in these buttons can be clicked, which opens a menu to edit the respective filter value.
* Filters can be removed by clicking the "x". When filtering by name, deleting the query also removes that filter.
* When filtering by date, the workflow is different from the old UI. When a filter by date is first created, only a date picker is shown but no time of day can be specified. Once a date filter has been added, the filter button can be clicked to refine the time.
* Filters that use Inputs do not update the filters as soon as the value is changed by typing. The new value has to submitted by hitting enter. **A follow-up issue** (DIR-705) has been created to add a submit button to the input as soon as the respective UI component has been implemented.

## To do

* Investigate why Playwright tests often fail in the CI for this branch. May or may not be related to my changes.

## Notes / Feedback requested

* Note: There is a bug in the API. Before and after do the opposite of what you would expect. This should already be fixed, but not in the version on the frontend test server.

Here are some specific things I would like to receive feedback:

* When filtering by status and trigger, the value displayed on a defined filter button is not translated, but the string that is actually used in the query is rendered. This is apparent when the option "cloud event" (translated) is clicked, which is displayed as "cloudevent". It should be possible to translate this value (with some extra code required due to typescript), but it could be argued that it is actually better to show the exact value to users that they would also find in other places (for example, in the logs, or when using a cli tool). In that case, one could even argue some of the other labels in the menus shouldn't be translated either.
* Filter field "Trigger" was "Invoker" in the old UI, but is "TRIGGER" in the API. I think trigger is easier to understand in any case and we shouldn't use different labels to avoid confusion.
* I thought about componentizing the filter buttons further to make the code drier. But did not see easy ways to do so (without leading to problems with TypeScript due to merged types in JSX return blocks).
* The Filters component has gotten a bit long for my taste. It might be possible to separate the code into smaller components (for example, the filter buttons for existing filters could be separated from the main menu), but I hesitated to do this as it would involve a lot of prop drilling (or maybe introducing useContext()). The reason for this is that all the popover menus need to be handled in the main component's scope. Any suggestions on how to approach this are welcome (or maybe the code is acceptable for now).


src/pages/namespace/Instances/List/index.tsx
```diff
@@ -37,7 +40,8 @@ const InstancesListPage = () => {
         {t("pages.instances.list.title")}
       </h3>
       <Card>
-        <Table>
+        <Filters filters={filters} onUpdate={setFilters} />
+        <Table className="border-t">
```
stefan-kracht, line 44: I just noticed that the border color in dark mode is not correct (I guess it should match the table's outline)

This should fix it (I would also set the color explicitly for light mode as well): `<Table className="border-t border-gray-5 dark:border-gray-dark-5">`

 CleanShot 2023-07-25 at 13 35 31

src/pages/namespace/Instances/components/Filters/index.tsx
```diff
@@ -0,0 +1,265 @@
...
+              </Button>
+            </ButtonBar>
+          );
+        }
+
+        if (field === "AFTER" || field == "BEFORE") {
+          const dateValue = filters[field]?.value;
+          if (!dateValue) {
+            console.error("Early return: dateValue is not defined");
+            return <></>;
```
stefan-kracht, line 157: Could we just return `null` here? Even if the result is the same, I tend to use null over empty fragments when I can (and TypeScript doesn't complain here, I just tried it) because I find it semantically a bit more suitable to what it actually does. But it's just a suggestion, you can keep it that way, if you like.

src/api/instances/query/get.ts
```diff
@@ -6,39 +6,110 @@ import { instanceKeys } from "..";
...
     baseUrl,
     limit,
     offset,
+    filter,
   }: {
     baseUrl?: string;
     namespace: string;
     limit: number;
     offset: number;
+    filter?: string;
```
stefan-kracht, line 86: Just a thought that came into my mind. What do you think about changing this type from `string` to `FiltersObj` and use `getFilterQuery` inside the `getInstances`. This way `getInstances` would expose a more verbose type and all its consumers don't have to care about how to generate this string.

stefan-kracht (changes requested):
This is very well done! The new UI is a huge improvement and the code is super clean and has a very good abstraction. Great work @sebxian!!

*bonus points for the unit test 🥳

## regarding your comments

> When filtering by status and trigger, the value displayed on a defined filter button is not translated, but the string that is actually used in the query is rendered. This is apparent when the option "cloud event" (translated) is clicked, which is displayed as "cloudevent". It should be possible to translate this value (with some extra code required due to typescript), but it could be argued that it is actually better to show the exact value to users that they would also find in other places (for example, in the logs, or when using a cli tool). 

I think it's okay to show it as cloudevent instead of cloud event. It is the same format as it will be shown in the table below as well. So that will be consistent then. 

> In that case, one could even argue some of the other labels in the menus shouldn't be translated either.

Do you mean in the filter dropdown? Good question, it should probably be the same (like e.g. cloudevent) I think.

> Filter field "Trigger" was "Invoker" in the old UI, but is "TRIGGER" in the API. I think trigger is easier to understand in any case and we shouldn't use different labels to avoid confusion.

I agree that it should be consistent. Also with the table header, which currently shows "invoker". That's also the property name from the instance's api response. I wonder which term is  "more correct"  😀

> The Filters component has gotten a bit long for my taste. It might be possible to separate the code into smaller components (for example, the filter buttons for existing filters could be separated from the main menu), but I hesitated to do this as it would involve a lot of prop drilling (or maybe introducing useContext()). The reason for this is that all the popover menus need to be handled in the main component's scope. Any suggestions on how to approach this are welcome (or maybe the code is acceptable for now).

I would accept this for now, but we can improve this in the future. I think I would prefer to use zustand, because we have the library included anyway (currently always with the persist middleware, which we don't need here) and to me the code to implement the same logic is much cleaner and more readable in comparison to react context.

## functional review
- I think we should reset the pagination when the filter changes. This can currently lead to a bug that shows no results, even if there are some. To reproduce:
  - trigger 1 parent that spawns 30 child instances
  - go to instances page (1 instance has the trigger `api`, 30 instances have the trigger `instance`)
  - go to page 2
  - filter by trigger `api`
  - no result is showing (because we are still on page 2 and don't see it)
- when a filter was applied and we don't see any results, it shows the "there are no recently executed instances" text. Should we add another "no search results"-like text for this case or make the default message more generic?
- when a filter was applied, could we remove it from the context menu that appears after clicking the + button? When I construct a complex filter, I think it would be useful to not see the filters you already set. What do you think?

## design review
- the first part of the filter is currently a button, and I think we probably should make it a label? This might be an extra ticket for aaron to make our ButtonBar component support labels.
CleanShot 2023-07-25 at 13 47 02
-  I was also wondering if we should visually separate the filter values to give the filter bar a bit more structure. Just a very quick attempt  (added `className="font-normal text-gray-11"` to the button)
CleanShot 2023-07-25 at 14 04 13

src/pages/namespace/Instances/components/Filters/index.tsx
sebxian: `null` makes a lot more sense, thank you! I was unhappy about this, but wasn't aware this is possible. Fixed in 046073f86c339fde8545b04bba646957f79ea3a0

src/api/instances/query/get.ts
stefan-kracht: For the query key you should be able to [pass the object](https://tanstack.com/query/v4/docs/react/guides/query-keys#query-keys-are-hashed-deterministically) as it is without needing to call `getFilterQuery` at all. As far as I understand, every change in the filterobject will generate a different filter query string. 

We just have to be aware that
`const filterObj = { "key" : "value", anotherKey: undefined}` and `const filterObj = { "key" : "value" } ` would share the same cache key. But that should not be a problem.

sebxian:
@stefan-kracht regarding consistent terms, I updated the table header "invoker" to "trigger" in bf2d4aabe243c1d4181e77367334bb82218b1178

I didn't update variable names in the backend because I realized that the inconsistent naming is actually in the API. The property in the "instances" model is called "invoker", but the filter param is called "TRIGGER". 

In 5ef0729c6a5165ef667dec29c7997a201110567e, I removed the translations for the enum filter values in the dropdown menu (those are used under trigger and status). Now they are 100% identical with what will be shown on the filter button (but there were only whitespace differences anyway). It did look a bit better before and we could always revert this if that's a problem.

sebxian:
@stefan-kracht I reset the offset when updating filters here: 03974c148ee89ddebd4e66f629a8421ce956e2a0

Good catch, that's something I normally do, but forgot this time.

src/pages/namespace/Instances/List/index.tsx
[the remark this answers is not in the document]
sebxian: applied in bf2d4aabe243c1d4181e77367334bb82218b1178

sebxian:
@stefan-kracht I followed all your suggestions (except one noted below), please go through my commits and let me know what you think!

As discussed in person, I didn't update the styling for the filter buttons to leave that for a separate issue.

Also, I'm not sure I like removing the filters from the menu when they have been defined. If they are always (and redundantly) in the + button menu, some users may find it easier to find a filter via this menu and change the value than find the correct filter button. I think I prefer the redundant version (but I don't feel strongly about it).

stefan-kracht (approved):

PR #422, opened 2023-07-20
[DIR-651] Instance List E2E Test

porleaaron:
### What's done

1. Test Instance List item without pagination
2. Test pagination
3. Test child instance
###  Need to discuss

1. How to test the content of a certain page
2. How to test the different buttons of the pagination component


e2e/instances/list/index.spec.ts
```diff
@@ -0,0 +1,388 @@
...
+  ).toContainText(instanceDetail?.status.toString() || "pending");
+
+  if (instanceDetail?.status === "failed") {
+    await state.hover();
+    const errorTooltip = page.getByTestId(
+      `instance-row-state-error-tooltip-${instance.instance}`
+    );
+    await expect(
+      errorTooltip,
+      "on hover the failed badge, error tooltip should appear"
```
stefan-kracht, line 176: I think we can test against the exact text of this error message

e2e/instances/list/index.spec.ts
```diff
@@ -0,0 +1,388 @@
...
+  ).toHaveURL(`/${namespace}/instances/${instance.instance}`);
+
+  //no pagination to be visibe
+  await expect(
+    page.getByTestId("instance-list-pagination"),
+    "there is no pagination when there aren't many instances"
+  ).not.toBeVisible();
+});
+
+test("test the pagination", async ({ page }) => {
```
stefan-kracht, line 239: About this test in general, can you go over it and check if you can introduce some more variables that make the numbers more understandable. You could for example give `Math.floor(Math.random() * 12)` a more meaningful name. Even the numbers, `12`, `170` and `15` could be variables to make it easier to understand the context.

e2e/instances/list/index.spec.ts
```diff
@@ -0,0 +1,388 @@
...
+          limit: 15,
+          offset: (p - 1) * 15,
+        },
+      });
+      const firstInstance = instancesListOfPage.instances.results[0];
+      const instanceItemId = page.getByTestId(
+        `instance-row-id-${firstInstance?.id}`
+      );
+      await expect(instanceItemId, "ItemId should have the id").toContainText(
+        firstInstance?.id.slice(0, 8) || ""
```
stefan-kracht, line 331: I think this is an invalid test too. There is a scenario where you allow the instance id to be empty and the test to be valid. I think I understand why you are doing this, because `firstInstance.id.slice(0, 8)` (without the `?`) result in a typescript error, right? You can get around this by doing this

```
const firstInstance = instancesListOfPage.instances.results[0];
if (!firstInstance)
        throw new Error("there should be at least one instance");
```

this will make the test fail very early when the `firstInstance` is undefined and after this block, TypeScript will know that firstInstance can not be undefined at this point

e2e/instances/list/index.spec.ts
```diff
@@ -0,0 +1,388 @@
...
+  await expect(
+    page.getByTestId("pagination-wrapper"),
+    "there should be pagination component"
+  ).toBeVisible();
+
+  const btnPrev = page.getByTestId("pagination-btn-left");
+  const btnNext = page.getByTestId("pagination-btn-right");
+
+  //loop through all pages by clicking Next button
+  const testPages = [
```
stefan-kracht, line 281: I think we should be very careful with randomness here. With this approach, a failing test is very hard to reproduce. If we have e.g. an issue on the last page, it will only be caught with a very low probability and only on certain test runs. And I actually have "sometimes" have problems with this test. 

CleanShot 2023-07-20 at 16 40 13@2x

Did you expierence any similar problems?

Maybe there was a misunderstanding in what we need to test. I think we should go through every page and check the state of the pagination on each one.

stefan-kracht (changes requested):
Thanks, @porleaaron! I added some comments with suggestions directly in the files.

Some general thoughts:

- I think the workflow name in the row is not tested yet.
- Can you add one more test to check the behavior of a child instance? It can be one new test where a parent creates one child and then just test the things I mentioned in the ticket

> Testcases for a child instance (please ask me about a test workflow for that, I added an example below):
> revision id shows "none"
> invoker shows "instance", id reveals on hover

e2e/instances/list/index.spec.ts
stefan-kracht: I just spoke with Sebastian and discussed this problem with using random pages and that he suggested this approach this morning. I kind of missed that, sorry! 

We actually had a misunderstanding of what should be tested and now agreed on keeping it simple and make much less test on the pagination itself (we have unit tests in place for the general behavior of the pagination component). So please just generate instances for 2 and a half pages (it would be good to make the last page not completely filled with instances) and click through each of the pages and check the pagination and first (maybe also the last) item on the page.

I think this is a lot simpler to read as well, because we don't need to loop over all pages dynamically and rather just statically go to page 2 and 3.

e2e/instances/list/index.spec.ts
```diff
@@ -0,0 +1,388 @@
...
+  // const failedInstanceServerRes = instancesList.instances.results.find(
+  //   (x) => x.id === failedInstance.instance
+  // );
+
+  // console.log("🚀", failedInstanceServerRes);
+
+  // { waitUntil: "networkidle" } might not be necessary, I just added
+  // it so that running this example will show a nicely loaded page
+  await page.goto(`${namespace}/instances/`, { waitUntil: "networkidle" });
+});
```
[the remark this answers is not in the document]
porleaaron: Yes, should be removed

e2e/instances/list/index.spec.ts
```diff
@@ -0,0 +1,388 @@
...
+  const invoker = page.getByTestId(
+    `instance-row-invoker-${childInstanceDetail?.id}`
+  );
+  await expect(
+    invoker,
+    "invoker should appear in the row, to be instance for this child instance"
+  ).toContainText("instance");
+
+  // click on button 2 and check if the result is correct
+  // click on prev button and check if the result is correct
```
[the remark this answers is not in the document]
porleaaron: I planned to test the number buttons of the pagination, but now I do it on randomly selected pages
So this should be removed

e2e/instances/list/index.spec.ts
```diff
@@ -0,0 +1,388 @@
...
+  // );
+
+  // console.log("🚀", failedInstanceServerRes);
+
+  // { waitUntil: "networkidle" } might not be necessary, I just added
+  // it so that running this example will show a nicely loaded page
+  await page.goto(`${namespace}/instances/`, { waitUntil: "networkidle" });
+});
+
+test("there is no result", async ({ page }) => {
```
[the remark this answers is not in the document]
porleaaron: yeah, I should update it

stefan-kracht:
I also committed some slight changes, I think the commit messages should explain everything.

porleaaron:
Those comments are all implemented

stefan-kracht (approved):

stefan-kracht:
Looks good @porleaaron, thank you for your quick changes! I made some small cleanup, I hope you are okay with that. Now, we are just in time to merge it before the weekend 🥳

porleaaron:
Gone through your commits
Thanks for those updates

PR #423, opened 2023-07-21
[DIR-698] diagram e2e tests

stefan-kracht:
This PR adds e2e tests for the diagram editor.

It was a bit tricky to test that the diagram updates after changing the workflow for various reasons:
- the workflow content must be changed via the editor to make it a valid test
- the editor is smart about code indention, when I tried to reuse one of our workflow templates, I would have to parse out various indents, which would make the test unreadable
- I also wanted to avoid/minimize the simulation of typing text 
- [insertText](https://playwright.dev/docs/api/class-keyboard#keyboard-insert-text) did not work because our editor is not a native textarea and relies on several events that are fired during typing, these events are bypassed with insertText, so the editor does not update
- using the clipboard to copy and paste the editor content would be my preferred solution, but it was not reliable
- my final solution was to use keyboard input but keeping it as minimal as possible by just commenting out various states

### DIFF
--- a/e2e/explorer/workflow/diagramEditor.spec.ts
+++ b/e2e/explorer/workflow/diagramEditor.spec.ts
@@ -1,7 +1,10 @@
 import { Page, expect, test } from "@playwright/test";
+import {
+  noop as basicWorkflow,
+  validate as complexWorkflow,
+} from "~/pages/namespace/Explorer/Tree/NewWorkflow/templates";
 import { createNamespace, deleteNamespace } from "../../utils/namespace";
 
-import { noop as basicWorkflow } from "~/pages/namespace/Explorer/Tree/NewWorkflow/templates";
 import { createWorkflow } from "~/api/tree/mutate/createWorkflow";
 import { faker } from "@faker-js/faker";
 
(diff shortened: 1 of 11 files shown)


sebxian (approved):
Good work. The tests are very clearly written and seem to work well :rocket: 

Commenting out the code seems like a clever solution under the circumstances.

I looked at the failed playwright job in the CI, but that seems to be unrelated to your changes. So this looks good to merge for me.

e2e/explorer/workflow/diagramEditor.spec.ts
```diff
@@ -0,0 +1,295 @@
...
+
+import { consumeEvent as consumeEventWorkflow } from "~/pages/namespace/Explorer/Tree/NewWorkflow/templates";
+import { createRevision } from "~/api/tree/mutate/createRevision";
+import { createWorkflow } from "~/api/tree/mutate/createWorkflow";
+import { faker } from "@faker-js/faker";
+
+let namespace = "";
+let workflow = "";
+
+const getCommonPageElements = async (page: Page) => {
```
sebxian, line 12: All tests work on my system if I remove the async and awaits.

From what I've understood from the PW docs, these selectors are executed when you use them, so that's only later in the code and there is nothing to wait for at this point.

PR #424, opened 2023-07-21
[DIR-703] input with button

porleaaron:
### What's done

1. Basic Wrapper for Input with Button

### Need to discuss

1.  Inputs is hidden behind the Button, should the input end before the Button starts?
At the end of this video, you can see the current behavior 
https://www.loom.com/share/1edfa9d131744e94b274a247ef3ca1a8

### DIFF
--- a/src/design/InputWithButton/index.stories.tsx
+++ b/src/design/InputWithButton/index.stories.tsx
@@ -1,4 +1,3 @@
-
 import type { Meta, StoryObj } from "@storybook/react";
 import {
   Tooltip,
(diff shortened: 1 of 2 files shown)


stefan-kracht:
I have just seen, that this is a draft PR and you commented on it. I already wanted to review it and pushed some stoybook changes. Just small changes to make the stories actually work.

Regarding your question:
Yes, the design is good already, but the text should stop before the button.

stefan-kracht:
> Yes, that was the problem, to stop the text before the Button starts
> 
> 1. Icon buttons have fixed width so we can set padding for text input
> 2. The textbutton has dynamic width, so that makes it difficult to set the padding for text input
> 
> The discussion point should be how to set dynamic padding of the input according to the dynamic width of the text button

For simplicity, let's remove the text button for now and only focus on a `<Button icon />` to be compatible. This way the InputWithButton component should be much simpler and should not need any prop.

stefan-kracht:
Good job @porleaaron

PR #425, opened 2023-07-26
Use global teardown to delete e2e test  namespaces

sebxian:
## Description

This removes the `deleteNamespace` from individual tests (and the afterEach hook respectively), and instead deletes all namespaces created during the test run in a [global teardown](https://playwright.dev/docs/test-global-setup-teardown) method.

Since deleting namespaces is expensive, postponing it until all tests have finished should improve the speed / stability of the tests.

The first jobs still had flaky tests (mainly concerned with the workflow editor and diagram?), but seemed to be a bit faster than previous jobs.

I'll leave this on "draft" for now until I have time to compare it to other test runs, and maybe retrigger it a few times to get a better impression of how it affects the tests.

### DIFF
--- a/e2e/explorer/index.spec.ts
+++ b/e2e/explorer/index.spec.ts
@@ -2,7 +2,6 @@ import {
   checkIfNamespaceExists,
   createNamespace,
   createNamespaceName,
-  deleteNamespace,
 } from "../utils/namespace";
 import {
   checkIfNodeExists,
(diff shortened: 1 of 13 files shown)


sebxian:
Closing this, mainly for two reasons:
* the concept seems valid but deleting namespaces in global teardown seems less reliable (so there is a risk for many left-over namespaces)
* tests currently work well so the potential performance gain doesn't seem to be too big an issue

sebxian:
Closing this, mainly for two reasons:
* the concept seems valid but deleting namespaces in global teardown seems less reliable (so there is a risk for many left-over namespaces)
* tests currently work well so the potential performance gain doesn't seem to be too big an issue

PR #427, opened 2023-08-02
[DIR-709] buttonbar needs to support label, update button component to have asLabel props

porleaaron:
### What's done

1. update the Button component to have props asLabel
2. update the ButtonBar component to handle the "label" children.
3. update the stories for Button and ButtonBar component
### What needs to be discussed

1. Does label have Hover effect as I have removed the click effect from button in case of asLabel
2. Any other UI requirements in case of label?

### DIFF
--- a/src/design/ButtonBar/index.stories.tsx
+++ b/src/design/ButtonBar/index.stories.tsx
@@ -107,10 +107,7 @@ export const AsChildButtons = () => (
     <p>No Click, Hover effects on child components except Anchor</p>
     <ButtonBar>
       <Button asChild size="lg">
-        <label>Label1</label>
-      </Button>
-      <Button asChild size="lg">
-        <label>Label2</label>
+        <label>Label</label>
       </Button>
       <Button asChild size="lg">
         <a href="/">Link</a>
@@ -123,9 +120,6 @@ export const AsChildButtons = () => (
       </Button>
       <Button asChild size="lg">
         <span>Span Tag</span>
-      </Button>
-      <Button variant="primary" asChild size="lg">
-        <span>No Hover</span>
       </Button>
     </ButtonBar>
   </div>
--- a/src/design/ButtonBar/index.stories.tsx
+++ b/src/design/ButtonBar/index.stories.tsx
@@ -149,11 +149,7 @@ TLDR; It could technically solved, but all state attributes would need to be
 namespaced which would have a DX impact that is not worth it for now.
              */}
 
-            {/* 
-                add !border-0 class in wrapper so trigger wouldn't have border for both button and its wrapper(div)
-            */}
-
-            <div className="!border-0">
+            <div className="">
               <Toggle aria-label="Toggle italic">
                 <HelpCircle />
               </Toggle>
@@ -169,7 +165,7 @@ namespaced which would have a DX impact that is not worth it for now.
         </Button>
         <Tooltip>
           <TooltipTrigger asChild>
-            <div className="!border-0">
+            <div className="">
               <Button variant="outline" aria-label="Toggle italic">
                 <Bug /> Button with tooltip
               </Button>
--- a/src/design/ButtonBar/index.stories.tsx
+++ b/src/design/ButtonBar/index.stories.tsx
@@ -149,7 +149,7 @@ TLDR; It could technically solved, but all state attributes would need to be
 namespaced which would have a DX impact that is not worth it for now.
              */}
 
-            <div className="">
+            <div>
               <Toggle aria-label="Toggle italic">
                 <HelpCircle />
               </Toggle>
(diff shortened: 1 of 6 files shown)


stefan-kracht:
EDIT: Sorry, that comment was on the wrong PR :) 

Can you elaborate on the status of this PR? I see an implementation of multiple approaches that we discussed. So does our preferred one work? 

```typescript
<Button asChild>
  <label>Label</label>
</Button>
``` 

Also the `asLabel`, was meant to be `as="label"` and could be anything else as well (like `as="div"`). That surely was not clear, because you haven't seen it written down :) But this would only be our fallback solution if the first one would not work.

stefan-kracht:
As soon as this is a draft PR it would be great if you give me some context of the status. It would be much easier for me to know what you would like me to check. E.g I see some TS errors and I don't know if they are intentional.

stefan-kracht:
The code in the Button Component looks very good to me!! If that is what you would like me to check :)

porleaaron:
This branch has now all we wanted at the beginning, you can have a review @stefan-kracht

src/design/ButtonBar/index.stories.tsx
```diff
@@ -141,7 +175,7 @@ namespaced which would have a DX impact that is not worth it for now.
         </Button>
         <Tooltip>
           <TooltipTrigger asChild>
-            <div>
+            <div className="!border-0">
```
stefan-kracht, line 178: We somehow have to find a way to get rid of this with a tailwind css selector in the ButtonBar Component? Removing the border should all be handled by the ButtonBar. 


I think this also broke this part of our workflow editor

CleanShot 2023-08-03 at 09 18 40@2x

src/design/ButtonBar/index.tsx
```diff
@@ -15,6 +15,15 @@ export const ButtonBar: React.FC<HTMLAttributes<HTMLDivElement>> = ({
       "[&_button]:active:outline-none",
       "[&_button]:border [&_button]:focus:ring-0 [&_button]:focus:ring-offset-0",
       "[&_button]:active:ring-0 [&_button]:active:ring-offset-0",
+
```
stefan-kracht, line 18: Can you clean up the comments here? There are some empty l ines and a 
`//in case of label`  comment that I don't get.

src/design/ButtonBar/index.stories.tsx
```diff
@@ -102,6 +102,35 @@ export const ButtonSizes = () => (
   </div>
 );
 
+export const AsChildButtons = () => (
+  <div className="flex flex-col gap-4">
+    <p>No Click, Hover effects on child components except Anchor</p>
+    <ButtonBar>
+      <Button asChild size="lg">
```
stefan-kracht, line 109: Could only one Label be enough here? I was just wondering what the difference of Label1 and Label2 is

src/design/ButtonBar/index.stories.tsx
```diff
@@ -102,6 +102,35 @@ export const ButtonSizes = () => (
...
+        <GitBranchIcon /> Review
+      </Button>
+      <Button size="lg">
+        <SearchIcon /> Search
+      </Button>
+      <Button asChild size="lg">
+        <span>Span Tag</span>
+      </Button>
+      <Button variant="primary" asChild size="lg">
+        <span>No Hover</span>
```
stefan-kracht, line 128: What does "No Hover" mean here

stefan-kracht (changes requested):
I added some comments and this part of the [Ticket](https://linear.app/direktiv/issue/DIR-709/buttonbar-needs-to-support-label) is still missing (the filters on the instances list need to be updated accordingly):
- [ ] change the buttons on every filter to be a label

src/design/ButtonBar/index.stories.tsx
porleaaron: Primary button has the Hover Effect 
But when we apply asChild props, and the child is Span, then there is no hover effect
Only <a> child will have the hover effect, click effect
This is what I wanted to say in the story

src/design/ButtonBar/index.stories.tsx
porleaaron: Yes, both labels are the same, we can remove one of them

src/design/ButtonBar/index.stories.tsx
```diff
@@ -125,7 +148,8 @@ https://github.com/radix-ui/primitives/discussions/560
 TLDR; It could technically solved, but all state attributes would need to be 
 namespaced which would have a DX impact that is not worth it for now.
              */}
-            <div>
+
```
stefan-kracht, line 151: can you please remove empty `className` attributes?

stefan-kracht:
Thanks @porleaaron, I quickly did this one task that was still missing

> I added some comments and this part of the [Ticket](https://linear.app/direktiv/issue/DIR-709/buttonbar-needs-to-support-label) is still missing (the filters on the instances list need to be updated accordingly):
> 
> * [ ]  change the buttons on every filter to be a label

stefan-kracht (approved):

PR #428, opened 2023-08-02
[DIR-712] - add inputWithButton stories for some forms

porleaaron:
### What's done

1. Fix styling of inputWithButton component
2. Add stories for some existing forms

### DIFF
--- a/src/componentsNext/NamespaceCreate/InfoTooltip.tsx
+++ b/src/componentsNext/NamespaceCreate/InfoTooltip.tsx
@@ -1,0 +1,22 @@
+import { FC, PropsWithChildren } from "react";
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipProvider,
+  TooltipTrigger,
+} from "~/design/Tooltip";
+
+import { Info } from "lucide-react";
+
+const InfoTooltip: FC<PropsWithChildren> = ({ children }) => (
+  <TooltipProvider>
+    <Tooltip delayDuration={100}>
+      <TooltipTrigger type="button">
+        <Info size={16} className="ml-2 text-gray-11" />
+      </TooltipTrigger>
+      <TooltipContent className="max-w-md text-left">{children}</TooltipContent>
+    </Tooltip>
+  </TooltipProvider>
+);
+
+export default InfoTooltip;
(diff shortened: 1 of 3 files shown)


stefan-kracht:
Can you elaborate on the status of this PR? I see an implementation of multiple approaches that we discussed. So does our preferred one work? 

```typescript
<Button asChild>
  <label>Label</label>
</Button>
``` 

Also the `asLabel`, was meant to be `as="label"` and could be anything else as well (like `as="div"`). That surely was not clear, because you haven't seen it written down :) But this would only be our fallback solution if the first one would not work.

sebxian (changes requested):
Thanks, @porleaaron !

I noticed a few things that were not clear to me from the start, so I couldn't give more detailed instructions. It would be great if you could look at the following two topics and make the respective changes:

### Storybook folder structure

* Good idea to add the InfoTooltip as a general component! That means we should also have a story for that component (so it can be found in the storybook list). Could you please add this?
* I just talked to Stefan and we decided we should have one storybook folder where we collect examples on how to combine components. So could you please do the following:
  * make a top-level folder called "Composition" in storybook
  * add a sub-level folder "Forms" and move the form examples here (they should not be under the InputWithButton).

### Button layout

I noticed the "submit" and "cancel" buttons are full width in the one form example you added (probably they are `layout: box` in CSS?).

In the real app, we have them side by side. In these cases, the form is wrapped in a `<Dialog>` and the `<DialogFooter>` controls the button layout. 

I am not sure how to solve this, but I think it would be good to have the same layout in storybook as well. So can you please find out:
* is it possible to use the `<Dialog>` or `<DialogContent>` component as a wrapper in storybook without a "real" popover dialog? In this case, the example would better reflect the real world use case.
* As an alternative, it is probably easier just to add a wrapper around the buttons that groups them next to each other (similar to the examples in the Buttons stories). This would also be a helpful example to have in storybook.

It would be great if you could experiment with these options and report back which one works best and makes the most sense. 

Let me know if there are any questions!

porleaaron:
Yeah, thanks for your feedback
I will work for those comments

sebxian:
Closing this. "Composition" stories for forms make sense, but more work is required to build forms that are a useful template for development (with working examples of all components we have developed so far).

PR #431, opened 2023-08-02
DIR-471 Add mirror form to namespace create dialog

sebxian:
## Description

* Added new tab to namespace create dialog that makes it possible to create mirror.
* Depending on the namespace type (mirror or simple namespace), and mirror auth method (none, ssh, token), the validation schema changes so only valid data can be submitted. 
* Updated the NodeSchema to be compatible with latest API changes (without this, rendering of non-workflow files in a mirror can fail). There is still work to be done on rendering non-workflow files correctly, but that should be in the scope of another ticket (DIR-475).

## Notes for testing and review

* Aaron has already created an InfoTooltip component under `src/componentsNext/NamespaceCreate/InfoTooltip.tsx` in https://github.com/direktiv/direktiv-ui/pull/428/, I think I will use that one when it has been merged and remove the one under `src/componentsNext/NamespaceCreate/InfoTooltip.tsx`.
* We briefly discussed whether a hint should be given when a url is entered and the required format changes between ssh/https. As soon as validation has been triggered once, this is now reflected in the validation errors that are updated when changing criteria. However, I could not conceptualize a good way to check the url field separately before the whole form is validated (mostly in terms of what would make sense UX wise. useForm does offer methods to trigger validation for individual fields, but I think it would be confusing to show only a single error when other fields are theoretically invalid as well, but just haven't been checked yet).
* With partially complete / valid inputs, switch between mirror and non-mirror and different auth types. As soon as validation has been triggered once (by submitting with an invalid form), the errors rendered should always appear/disappear according to the different validation criteria. 
* I have tested token based authentication with a private repository at https://github.com/xbow/direktiv-mirror.git. I will share the access token created for that repository on request.
* I have not tested ssh authentication with a real key pair, but the new UI should behave exactly like the old UI. Maybe it is good enough to ask the back-end colleagues to test that everything works correctly after merging.
* The back-end accepts invalid data without returning an error. For example. It will create a new namespace even when a wrong SSH key pair is entered (it will just not be populated with any files, I suspect that syncing just fails silently?). This is something we should improve in a later iteration (make the API more verbose regarding errors).

## Update 20230804

In 7a057d2dc5fa930e1a57432dcaf361fdac0b0718, I applied css fixes that Aaron added in https://github.com/direktiv/direktiv-ui/pull/428 to this branch. I think that other PR will need some discussion how Storybook should be organized, which I would like to postpone until we have met our milestones for the release; but we should already use these small css fixes now.

### DIFF
--- a/src/componentsNext/NamespaceCreate/index.tsx
+++ b/src/componentsNext/NamespaceCreate/index.tsx
@@ -28,6 +28,7 @@ import FormErrors from "~/componentsNext/FormErrors";
 import InfoTooltip from "./InfoTooltip";
 import Input from "~/design/Input";
 import { InputWithButton } from "~/design/InputWithButton";
+import { Textarea } from "~/design/TextArea";
 import { fileNameSchema } from "~/api/tree/schema";
 import { pages } from "~/util/router/pages";
 import { useCreateNamespace } from "~/api/namespaces/mutate/createNamespace";
@@ -257,7 +258,7 @@ const NamespaceCreate = ({ close }: { close: () => void }) => {
                     {t("components.namespaceCreate.label.token")}
                   </label>
                   <InputWithButton>
-                    <Input
+                    <Textarea
                       id="token"
                       data-testid="new-namespace-token"
                       placeholder={t(
@@ -276,13 +277,13 @@ const NamespaceCreate = ({ close }: { close: () => void }) => {
                 <>
                   <fieldset className="flex items-center gap-5">
                     <label
-                      className="w-[144px] overflow-hidden text-right text-[14px]"
+                      className="w-[90px] overflow-hidden text-right text-[14px]"
                       htmlFor="passphrase"
                     >
                       {t("components.namespaceCreate.label.passphrase")}
                     </label>
                     <InputWithButton>
-                      <Input
+                      <Textarea
                         id="passphrase"
                         data-testid="new-namespace-passphrase"
                         placeholder={t(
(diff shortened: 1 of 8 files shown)


src/componentsNext/NamespaceCreate/index.tsx
```diff
@@ -61,14 +102,32 @@ const NamespaceCreate = ({ close }: { close: () => void }) => {
...
   const disableSubmit = !isDirty || (isSubmitted && !isValid);
 
+  // if the form has errors, we need to re-validate when isMirror or authType
+  // has been changed, after useForm has updated the resolver.
+  useEffect(() => {
+    if (isSubmitted && !isValid) {
+      trigger();
+    }
+  }, [isMirror, authType, isSubmitted, isValid, trigger]);
+
```
stefan-kracht, line 130: I think you have already suspected this last week :D I'm not a fan of this effect, but I totally get why you needed it, and I just tried so solve this problem by implicitly calling the trigger function when the authType changes, but this obviously does not work because in that function scope the value of `authType` has not changed yet and the `trigger()` validates against the "old" schema.

I think this is good for now and a searching for a cleaner solution should not block us at the moment.

stefan-kracht (changes requested):
Very good @sebxian. Especially with the three different zod validators. You found a nice way to abstract that validation complexity away from the component and make it much more readable.

I have made some comments in the code, but nothing major. 

This might be a mac problem, but we should give the labels just a tiny bit more space

CleanShot 2023-08-07 at 08 40 29

sebxian:
Thanks @stefan-kracht, I made the requested changes.

Could you double check that the label width works under OSX? Under Linux, the font is narrower so I didn't have this problem, but I added enough pixels that it should be fine :-).

stefan-kracht:
> Could you double check that the label width works under OSX? Under Linux, the font is narrower so I didn't have this problem, but I added enough pixels that it should be fine :-).


👌
CleanShot 2023-08-07 at 14 11 02

Feel free to merge at any time

stefan-kracht (approved):

PR #432, opened 2023-08-03
[DIR-716] remove revision from the instance list

stefan-kracht:
As the api will no longer provide the revision at the instance endpoint (`as: "workflow:latest` changed to `as: "workflow`) we need to remove the column.

This also makes the code a lot cleaner since we now always link to the latest revision.

I have adjusted the e2e tests, but without running them. It would be great if @sebxian could run the tests locally against an api that doesn't require an api token.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,6 +7,11 @@
 # e2e tests will also run against this API
 VITE_DEV_API_DOMAIN="http://api.for.development.com"
 
+# If the above API requires an authentication token, the following must be set
+# in order to run e2e tests against that API. For normal use, this is not relevant,
+# as the API token will be entered via the UI and saved in local storage.
+# VITE_E2E_API_TOKEN="abcde"
+
 # only neccesary when building for production, will be displayed underneath the login
 VITE_APP_VERSION=""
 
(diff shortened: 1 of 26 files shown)


sebxian:
@stefan-kracht,

I updated the playwright version (the VSCode extension required this for feature compatibility) and that actually (a) seems to have made the tests a bit more stable and (b) makes error reporting more accurate when I run the tests in VSCode.

After yarning to apply the update, this also requires to perform `npx playwright install` again to download updated browsers.

I fixed some issues in the tests, now we are only left with one test assertion with the message `save button should be disabled during the api call`, this occurs in two tests. Do you have an idea what could be wrong?

One thing I wanted to note: It is very important we avoid using `faker.git.shortSha()` or any such thing for creating filenames in the future. The backend doesn't accept filenames that start with a number, which can randomly happen with shortSha(). In that case, the tests fail. I suspect that is one reason behind some of our flaky tests.

sebxian (approved):
Good work, @stefan-kracht ! 

Running the e2e tests, I found some small things that had to be fixed (see also my earlier comment).

More remaining issues with the e2e tests have been fixed in https://github.com/direktiv/direktiv-ui/pull/435. They were not related to revision_ids specifically, but to other API updates.

At this point, this looks good to merge for me. But please double-check this commit: 172d4fea9cedd06011c73df96a7d6cf04cdb166a

I am not completely happy with this handling of file paths (including slashes) in the url, but probably we should leave it alone for now (and find a better approach with the v2 API).

stefan-kracht:
Thank you Sebastian. I checked the test you updated, and it looks good to me. There was a change in this field on the api which requires to remove the / in that test.

stefan-kracht:
I tried to enable playwright in CI but it did not work right away (I have set the new env var in github). I will merge for now and will investigate this in a separate PR.

PR #435, opened 2023-08-03
Add API token authentication to playwright

sebxian:
## Description

This enables playwright to run against an API that requires an authentication token.

This is based on the branch `stefankracht/dir-716-remove-revision-id-from-the-frontend` because that already fixes some tests. So the PR is set up to merge into that branch.

What I did:
* Add headers to all requests originating in playwright (mostly for creating and cleaning up test data).
* Add a local storage context that will be applied to the front-end app when executed in playwright.
* Updated the .env.example (variable to set for using token in e2e tests)
* There is a comment in the new file `e2e/setup/auth.ts` explaining my approach to set up the browser context in playwright.

Notes:
* All tests ran without errors against the version deployed on the dev-server (please confirm during review that this also works from another machine against the dev-server).
* Locally, two individual tests failed (as reported in PR https://github.com/direktiv/direktiv-ui/pull/432), but there is no problem with the test setup (without authentication). 
* When no token is defined in env, the storage context is still used. But since the token field is empty in this case, no auth tokens are added to the actual requests. (If worked fine even before I added the guard, when it was `"direktiv-token": "undefined"`. So the API probably just ignores the field if it is not needed).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,6 +7,11 @@
 # e2e tests will also run against this API
 VITE_DEV_API_DOMAIN="http://api.for.development.com"
 
+# If the above API requires an authentication token, the following must be set
+# in order to run e2e tests against that API. For normal use, this is not relevant,
+# as the API token will be entered via the UI and saved in local storage.
+# VITE_E2E_API_TOKEN="abcde"
+
 # only neccesary when building for production, will be displayed underneath the login
 VITE_APP_VERSION=""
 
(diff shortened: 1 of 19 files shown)


stefan-kracht:
This is so cool (also the faker part) thanks for taking care of this. All test pass now. I will merge right away!

*I just found a [small typo](https://github.com/direktiv/direktiv-ui/pull/435/commits/e7c7d377a3262f96c7928d6d4f4668d9851396ea) in one of the comments

stefan-kracht (approved):

PR #437, opened 2023-08-04
[DIR-705] Add submit buttons to inputs in filters

sebxian:
:rocket: preview: https://direktiv-ui-refs-pull-437-merge.direktiv.dev/

### Changes
* Uses InputWithButton to add a submit button to the time input in the instance filters
* Ditto for the TextInput for filtering by name
* Fixes a serious bug I noticed by chance: The guard I had previously added to check that time values are not undefined also prevented them having a value of "0".

### DIFF
--- a/src/design/InputWithButton/index.stories.tsx
+++ b/src/design/InputWithButton/index.stories.tsx
@@ -32,10 +32,10 @@ export const Default: Story = {
 };
 
 export const IconWithToolTip = () => {
-  const [value, setValue] = useState("some value");
+  const [value, setValue] = useState("some very very looooooooooong value");
   return (
     <TooltipProvider>
-      <InputWithButton>
+      <InputWithButton className="w-60">
         <Input
           value={value}
           onChange={(e) => {
--- a/src/design/InputWithButton/index.tsx
+++ b/src/design/InputWithButton/index.tsx
@@ -14,7 +14,7 @@ export const InputWithButton: React.FC<HTMLAttributes<HTMLDivElement>> = ({
     <div
       {...props}
       className={twMergeClsx(
-        "&_input]:pr-10 relative flex w-full items-end",
+        "relative flex w-full items-end [&_input]:pr-10",
         className
       )}
     >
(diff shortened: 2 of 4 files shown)


stefan-kracht (changes requested):
I like this. Looks so much better with a button!

For me, the time input needs to be a tiny bit bigger now (might be a mac thing?).

CleanShot 2023-08-05 at 09 13 11@2x

I also wonder if we might use a different icon? My first impression of that checkmark was, that it looks like it passed a validation and not submit something (like e.g. the username you entered is available). [Maybe this](https://lucide.dev/icons/arrow-right)?

sebxian:
@stefan-kracht, thanks for the feedback. I updated the submit button icon to ArrowRight, as suggested (17671d312335b4008a794bae571b6fbc6de4bb34)

Regarding the width of the time input, I am not sure what's the best approach. There is no explicit width set on the element. The input uses `w-full` and the parent (Command component) uses `w-min`. Maybe changing it to (or adding) `min-w-min` would solve the issue? Since I cannot see the issue under Linux, I can't try out possible fixes. Could you maybe investigate a fix that works on your Mac?

stefan-kracht:
> @stefan-kracht, thanks for the feedback. I updated the submit button icon to ArrowRight, as suggested ([17671d3](https://github.com/direktiv/direktiv-ui/commit/17671d312335b4008a794bae571b6fbc6de4bb34))
> 
> Regarding the width of the time input, I am not sure what's the best approach. There is no explicit width set on the element. The input uses `w-full` and the parent (Command component) uses `w-min`. Maybe changing it to (or adding) `min-w-min` would solve the issue? Since I cannot see the issue under Linux, I can't try out possible fixes. Could you maybe investigate a fix that works on your Mac?

I merged the base branch into this branch and had to fix one tiny issue, that broke the layout. And I deleted an old proof of concept component that was not in used as well.

I can reproduce the time input issue outside our app. So it's a Mac/Chrome/Retina related bug. Let's just ignore that for now.

CleanShot 2023-08-07 at 15 15 38@2x

Feel free to merge 🚀

stefan-kracht (approved):

PR #439, opened 2023-08-07
[DIR-623] Delete namespace

stefan-kracht:
This PR adds the functionality to delete a namespace.

### [Preview 🚀](https://direktiv-ui-refs-pull-439-merge.direktiv.dev/)

https://github.com/direktiv/direktiv-ui/assets/121789579/cc8b99b6-5740-4bb6-9bba-d0d6a42e0ff8

Follow-up tickets:
- [E2e tests](https://linear.app/direktiv/issue/DIR-728/test-delete-namespace)


public/locales/en/translation.json
```diff
@@ -295,6 +295,15 @@
           "workflowVariable": "Workflow Variable"
         }
       },
+      "deleteNamespace": {
+        "title": "Delete Namespace",
+        "description": "This will permanently delete the current active namespace and all resources associated with it.",
+        "deleteBtn": "Delete Namespace",
```
sebxian, line 301:
```suggestion
        "title": "Delete namespace",
        "description": "This will permanently delete the current active namespace and all resources associated with it.",
        "deleteBtn": "Delete namespace",
```

src/api/namespaces/mutate/deleteNamespace.ts
```diff
@@ -0,0 +1,65 @@
...
+    `/api/namespaces/${namespace}?recursive=true`,
+  method: "DELETE",
+  schema: NamespaceDeletedSchema,
+});
+
+type ResolvedCreateNamespace = Awaited<ReturnType<typeof deleteNamespace>>;
+
+export const useDeleteNamespace = ({
+  onSuccess,
+}: { onSuccess?: (data: ResolvedCreateNamespace) => void } = {}) => {
```
sebxian, line 21:
```suggestion
type ResolvedDeleteNamespace = Awaited<ReturnType<typeof deleteNamespace>>;

export const useDeleteNamespace = ({
  onSuccess,
}: { onSuccess?: (data: ResolvedDeleteNamespace) => void } = {}) => {
```

src/api/namespaces/mutate/deleteNamespace.ts
```diff
@@ -0,0 +1,65 @@
...
+    onSuccess(data, variables) {
+      /**
+       * invalidating the cache is important here, because after deleting the namespace
+       * we will redirect to the frontpage, where we pick the first namespace we can
+       * find and redirect to it. It is very likely that the cache will be used here
+       * (namespace cache gets populated very ealy in the app lifecycle), so we need
+       * to make sure that we don't accidentally redirect to the namespace we just
+       * deleted.
+       */
+      queryClient.invalidateQueries(namespaceKeys.all(apiKey ?? undefined));
```
sebxian, line 44: I wasn't sure about the right approach here but I think I was able to clarify it for myself, and I think the current implementation is good. Just sharing my thoughts FYI:

* Would it be feasible and sufficient to just delete the specific namespace from cache that we have just deleted?
* On the other hand, I think "single source of truth" is a good principle to follow here, except...
* ...what if deleting a namespace takes longer (because it's a cascading process) and the API still lists the namespace at this point in time? Well, I tested this and it doesn't, so that seems to be fine.

src/pages/namespace/Settings/index.tsx
```diff
@@ -52,6 +53,10 @@ const SettingsPage: FC = () => {
         <Broadcasts />
       </section>
 
+      <section data-testid="broadcasts-section">
```
sebxian, line 56:
```suggestion
      <section data-testid="namespace-section">
```

sebxian (commented):
Very nice :+1: 

Finally I was able to create some of those useless namespaces created for testing :smile: 

I suggested some naming/spelling corrections in the code, but otherwise this is approved!

src/api/namespaces/mutate/deleteNamespace.ts
stefan-kracht: > Would it be feasible and sufficient to just delete the specific namespace from cache that we have just deleted?

Yes, this would work as well, I think I wanted to keep it simple because this one is harder to test and just flushing the cache altogether will make it a bit more predictable. "Single source of truth" is a nice framing for that!

> ...what if deleting a namespace takes longer (because it's a cascading process) and the API still lists the namespace at this point in time? Well, I tested this and it doesn't, so that seems to be fine.

- I'm pretty sure the namespace is (marked as) deleted when the api call resolves, and some background tassk will clean up after that.

stefan-kracht:
Thanks for the review, I updated the branch (and resolved one merge conflict) and will merge now

PR #440, opened 2023-08-07
[DIR-194] Monitoring Page

stefan-kracht:
# This PR introduces the monitoring page in the new design

**[Preview 🚀](https://direktiv-ui-refs-pull-440-merge.direktiv.dev/)**

CleanShot 2023-08-10 at 09 38 41@2x

## Other Changes

Playwright tests have been re-enabled on every Pull Request once again.

I replaced the [Bug](https://lucide.dev/icons/bug) icon with the [ActivitySquare](https://lucide.dev/icons/activity-square)

Used the `TooltipCopyBadge` on the invoker column on the instance page as well (only for child instance)

fixes a bug in the Button component where the following code will not behave like a link (button is not clickable because the class `pointer-events-none` was accidentally applied)

```typescript
<Button asChild>
  <Link to="...">Link</Link>
</Button>
``` 
<details><summary>On the instances Page, the full screen button on the diagram panel had no background</summary>

| Before | After |
|--------|--------|
| CleanShot 2023-08-09 at 15 52 39@2x | CleanShot 2023-08-09 at 15 52 47@2x | 
</details> 

<details><summary>Added a new RefreshButton component for better feedback when the user clicked a refresh button</summary>

CleanShot 2023-08-10 at 08 42 37

</details> 

<details><summary>I changed the pattern on how we consume the streaming data in our components</summary>
<p>

The old API looked like this:

```typescript
...
const InstancesPage = () => {
  const { t } = useTranslation();

  /**
   * This hook will subscribe to a SSE stream. It will setup a connection
   * on mount and cancel the connection on unmount. To avoid unnecessary
   * reconnects, make sure to place this hooks in a isolated a parent component
   * that will not rerender very often.
   */
  useNamespaceLogsStream();

  return (
    <div className="flex grow flex-col gap-y-4 p-5">
      <h3 className="flex items-center gap-x-2 font-bold">
        <ActivitySquare className="h-5" />
...
``` 

We offer a hook for streaming, but it comes with a separate handbook on how to use it. When you place that hook wrong, it will reconnect a lot. Using streaming should be more simple and the context should not matter at all. The new solution lets us place a streaming provider component where ever we want (we can even place it deep down in a tree, when one of the parents is affected by the streaming updates and needs to re-render)

The problem of the previous solution was, that it was dependent on where it was placed. Since it was a hook it was re-rendering a lot, when the component it was in has changed. My idea was to somehow memo-ize the hook to avoid unnecessary re-renderings. But the hook has to have a `useEffect` in it and I did not find a good solution to memo-ize an effect (I think it is simply not possible without violating the exhaustive-deps eslint rule). 

So the solution would be to place it in its own component and put a `memo()` around it to only re-render when the props change. The new usage looks like this.

```typescript
...
const InstancesPage = () => {
  const { t } = useTranslation();
  return (
    <div className="flex grow flex-col gap-y-4 p-5">
      <NamespaceLogsStreamingProvider />  {/* could also be  <NamespaceLogsStreamingProvider enabled={somevar} /> */}
       ...
    </div>
...
```
My first intuition was to have it like this `<NamespaceLogsStreamingProvider>...</NamespaceLogsStreamingProvider>` but this would also re-render too often, because children are just props as well and memo will re-render when the children change. And it would also be kind of wrong to make this a wrapping component, because it would look like it would do something for its children while it does not have any parent child dependency. An empty component like `<NamespaceLogsStreamingProvider />` looks a bit weird and looks like it should be a hook, but I did not find a way to replicate the same behavior in a hook. 

Here is a screen recording of my test case:

https://github.com/direktiv/direktiv-ui/assets/121789579/dde1fd01-e4de-4771-96f1-3012d91e0917

- I put the provider directly inside the log component (which re-renders a lot when new logs comes in)
- I made a toggle for the `enabled` attribute of `NamespaceLogsStreamingProvider` to demo that the component will re-render when it has to (when props change)
- the network tab and the console output in this video shows when a new stream connection is established and when it is closed
- The old method would have shown a new connection in the network tab whenever we click that toggle, and also whenever a new log comes in. 
- the new method is able to avoid re-rendering when anything around it does, but it will rerender when it has to (e.g. when its own prop changes)
- I wonder if we should call it "provider". My mental model of a provider component would provide something to its children. What do you think about the name?
</p>
</details> 

## How to test

A good way to test the logs is to have two windows open (use two different browsers just to make sure you don't hit the browser limit for maximal number of open SSE connections) and watch the monitoring page as you trigger a workflow that spawns a lot of children automatically

## follow-up tickets
- [E2E test](https://linear.app/direktiv/issue/DIR-735/test-monitoring-page)


public/locales/en/translation.json
```diff
@@ -173,6 +173,30 @@
...
+        "tooltips": {
+          "copy": "Copy logs"
+        }
+      },
+      "instances": {
+        "openWorkflowTooltip": "click to open workflow {{name}}",
+        "updateTooltip": "refetch instances",
+        "successfullExecutions": {
+          "title": "Successful Executions",
+          "empty": "There are no recently successful executions."
```
sebxian, line 192:
```suggestion
          "empty": "There are no recent successful executions."
```
This wasn't strictly wrong, but a bit confusing. (recently successful = the success happened recently; recent successful executions = successful executions that occurred recently)

src/pages/namespace/Monitoring/Instances/Row.tsx
```diff
@@ -0,0 +1,113 @@
...
+          </ConditionalWrapper>
+        </TableCell>
+        <TableCell className="w-0 pr-5">
+          <Tooltip>
+            <TooltipTrigger>
+              {t("pages.instances.list.tableRow.realtiveTime", {
+                relativeTime: updatedAt,
+              })}
+            </TooltipTrigger>
+            <TooltipContent>{instance.updatedAt}</TooltipContent>
```
sebxian, line 107: I think we discussed this in the context of PR #442, and agreed to have a more human readable format in these cases.

We just need to decide if we quickly add it in the respective PRs or make a separate issue? The latter seems like almost a bit too much overhead, but maybe it's better because we could then fix all cases of this consistently.

src/pages/namespace/Monitoring/Instances/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+                  onClick={() => {
+                    refetchSucessfullInstances();
+                  }}
+                />
+              </TooltipTrigger>
+              <TooltipContent>
+                {t(`pages.monitoring.instances.updateTooltip`)}
+              </TooltipContent>
+            </Tooltip>
+          </TooltipProvider>
```
sebxian, line 76: Could we declare this as a `const refetchButton` or similar before the return statement and just reference it here? I think that would be more readable, because less nested.

src/pages/namespace/Monitoring/Logs/Scrollcontainer.tsx
```diff
@@ -0,0 +1,145 @@
...
+    <Logs
+      wordWrap={wordWrap}
+      className="h-full overflow-scroll"
+      ref={parentRef}
+      onScroll={(e) => {
+        const element = e.target as HTMLDivElement;
+        if (element) {
+          const { scrollHeight, scrollTop, clientHeight } = element;
+          const scrollDistanceToBottom =
+            scrollHeight - scrollTop - clientHeight;
```
sebxian, line 85: I'll mention this in the main PR review comment: There's a bug with scrolling to bottom in Firefox. 

I may be off target, but I have a hunch that somewhere around here something behaves differently in Firefox causing that bug, maybe due to the way it calculates these values?

Speaking of which, I think we should make a ticket for later to refactor the two occurrences of this `Scrollcontainer` into one reusable component.

src/pages/namespace/Monitoring/Logs/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { formatLogTime } from "~/util/helpers";
+import { useNamespacelogs } from "~/api/namespaces/query/logs";
+import { useTranslation } from "react-i18next";
+
+const LogsPanel = () => {
+  const { t } = useTranslation();
+  const { data } = useNamespacelogs();
+
+  const copyValue =
+    data?.results.map((x) => `${formatLogTime(x.t)} ${x.msg}`).join("\n") ?? "";
```
sebxian, line 21: Could "x" be named more clearly, for example "entry" if I understand correctly?

src/pages/namespace/Monitoring/Logs/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+              <TooltipContent>
+                {t("pages.monitoring.logs.tooltips.copy")}
+              </TooltipContent>
+            </Tooltip>
+          </TooltipProvider>
+        </ButtonBar>
+      </div>
+      <ScrollContainer />
+      <div className="flex items-center justify-center pt-2 text-sm text-gray-11">
+        <Loader2 className="h-3 animate-spin" />
```
sebxian, line 56: I don't have an obvious solution for this and it is ok to leave it as is for now. But intuitively, the spinning circle feels like something is wrong (as if there is content that should finish loading, but it's stuck). 

I am thinking that something like three pulsating dots (like a typing notification in a chat) could make sense here. I don't know if something like this exists in Lucide, but maybe the [ping](https://tailwindcss.com/docs/animation#ping) or pulse notification with a suitable icon would work?

sebxian (changes requested):
Very nice work, looks good and works very smoothly :rocket:

I left some annotations in the code, mainly regarding naming/readability (as usual from me :wink:) and nothing serious or urgent. Some of them may be worth postponing and solving in later issues.

Regarding your pattern with the streaming "providers", I think that is an excellent solution (that we could also apply in other cases, maybe we should make yet another issue to track this). However, I left a comment in the code, suggesting a naming change. 

One issue we should solve before release, in my opinion:
* Bug on Firefox (see video): when I scroll to the bottom of the logs (via scrollbar or mousewheel), it jumps to the top again. When I click "follow logs" to scroll down, it jumps to the top instead. (Maybe in both cases the jump to top is triggered by reaching the bottom?) 
[Screencast from 14.08.2023 10:09:54.webm](https://github.com/direktiv/direktiv-ui/assets/129740294/02b9d9ae-5077-4518-862f-e3707536639b). This also happens in the Instances log. I'm ok with making a linear ticket if we cannot solve it right now, but it should be fixed for the release in my opinion.

(I am a bit worried there may be further issues we're not aware of in Firefox and Safari)

Lots of potential separate issues, but as a said, it looks very good overall and except for the Firefox bug, I think none of the suggested changes are urgent.

public/locales/en/translation.json
stefan-kracht: Good catch. I was just copying from the old design...

src/api/namespaces/schema.ts
```diff
@@ -15,6 +16,29 @@ export const NamespaceCreatedSchema = z.object({
   namespace: NamespaceSchema,
 });
 
+export const NamespaceLogSchema = z.object({
```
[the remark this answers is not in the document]
stefan-kracht: I forgot to type comment (I just did that, it can be e.g "instance" 78b9b0c6a93ba03ab4e5982f336ffd9ae9875449). 

The thing is, we don't even need anything within the tags atm, we could even delete this validation node and it would have no effect.

But this brings up an interesting question about how we deal with fields that we don't need right now. Should we not include them in the schema at all (and risk possible future validation issues for fields that we don't even use) or should we keep them as part of our "documentation". I mean they could become handy when we quickly introduce a new feature with a field that we never used before. 


---
Maybe @h8718 can help us out here? It's about the result of a namespace log (like `/api/namespaces/NAMESPACE/logs`)

- what does "source" refer to? It is something like `c75454f2-3790-4f36-a1a2-22ca8a4f8020`
- "callpath" seems to be redundant with "instanceId" any idea what the point of it is?

src/design/Button/index.tsx
```diff
@@ -45,7 +56,9 @@ const Button = React.forwardRef<
   ) => {
     const Comp = asChild ? Slot : "button";
     const isAnchor = React.Children.toArray(children).some(
-      (child) => React.isValidElement(child) && child.type === "a"
+      (child) =>
```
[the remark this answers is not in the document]
stefan-kracht: We actually have a story for that: https://feature-redesign--6418690f3c0c39f7a2975666.chromatic.com/?path=/story/components-button--as-child

src/pages/namespace/Instances/Detail/index.tsx
```diff
@@ -4,29 +4,21 @@ import {
...
-   * will ensure that the hook will not rerun itself (canceling
-   * the stream and restarting a new one) when it updates the cache
-   */
-  useInstanceDetailsStream({ instanceId });
-  useLogsStream({ instanceId, filters });
-
-  return <InstancesDetail />;
+  return (
+    <>
+      <InstanceStreamingProvider instanceId={instanceId} />
```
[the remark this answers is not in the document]
stefan-kracht: This makes a ton of sense. I love that name! I just changed them all (858f97b60ad9d0ec0f82795922e9cf28a6fb4bb0)

src/pages/namespace/Instances/List/Row.tsx
```diff
@@ -80,33 +82,22 @@ const InstanceTableRow: FC<{
...
-                      onClick: (e) => {
-                        e.stopPropagation();
-                      },
-                    }}
-                  />
-                </TooltipContent>
-              </Tooltip>
-            )}
-          >
+          {isChild ? (
```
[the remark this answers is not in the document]
stefan-kracht: You are right, `isChild` can be misunderstood (fixed in 83b2ec6e033742df08ca80b519785b4719b82997)

src/api/namespaces/schema.ts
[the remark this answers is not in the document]
sebxian: > Should we not include them in the schema at all (and risk possible future validation issues for fields that we don't even use) or should we keep them as part of our "documentation".

Yes, good point, I am not sure about that myself. In a way it would be more efficient and straightforward to avoid definitions we never use. In some way they are "dead code", except they add dependencies that could break when the API changes. On the other hand, I do like this "documentation layer".

I think I have a slight preference for not validating props that we don't use, but I'm fine either way.

If we decide only to declare what we really use, we could add comments on top of the schema files with examples of back-end responses [like I did here](https://github.com/direktiv/direktiv-ui/blob/84afca459821f7307e2d9d14ad37c4584efa4f50/src/api/eventListeners/schema.ts). What do you think about that?

src/pages/namespace/Monitoring/Logs/index.tsx
stefan-kracht: Yes, I changed it to `logEntry` now (also in a similar code snippet in the instance logs) and will try to come up with better names for these inline function parameter :)  ca1fb2624402bfa153b6dd4c47eaa6dbc889511a

src/pages/namespace/Monitoring/Logs/index.tsx
stefan-kracht: I know what you mean. I tried the "someone is writing" kind of animation at first when I created the instances page and I agree that it feels like it is stuck in a loading state even if it is just a long-running stream with no data coming in.

The ping is nice! I've implemented it for both log streaming pages

CleanShot 2023-08-14 at 15 25 59 (e8d928e84367333baa3db9f13a34c502e2a7bfae)

src/pages/namespace/Monitoring/Instances/index.tsx
stefan-kracht: Sure! (89b0c95217d9070867419c1d149783d7ed36752d)

src/pages/namespace/Monitoring/Instances/Row.tsx
stefan-kracht: I think I would go for a separate issue, just because it's hard to solve this in all the open PRs at the same time.

src/pages/namespace/Monitoring/Instances/Row.tsx
```diff
@@ -0,0 +1,113 @@
...
+        className="cursor-pointer"
+      >
+        <TableCell className="grid pl-5">
+          <Tooltip>
+            <TooltipTrigger asChild>
+              <Link
+                onClick={(e) => {
+                  e.stopPropagation(); // prevent the onClick on the row from firing when clicking the workflow link
+                }}
+                to={pages.explorer.createHref({
```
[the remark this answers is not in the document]
sebxian: That's fine. Maybe I'll come back to this later :wink:

src/api/namespaces/schema.ts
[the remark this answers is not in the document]
h8718: `source` is the `source` where the log message originated from. Since at the moment the backend has a api-point for each `source-type` you don't need to care about that it does not have to be in the schema. But if its in its also not a problem since this value will be present anyway even if we completely redo the logging backend.  

We use it internally to filter the logs and return only relevant log messages.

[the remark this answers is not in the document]
stefan-kracht: Thanks for clarifiying @h8718!


As just discussed, I removed the unused attribute from our schema and added an example response as a comment (becf6637473f6607365a35a5104b910fd19a77a6)

stefan-kracht:
@sebxian Thanks you @sebxian, very helpful input as usual. I should have resolved all the open issues and will wait for your final approval 🚀

PR #441, opened 2023-08-08
Tooltip Copy Badge

sebxian:
This adds a design component and corresponding storybook story, as discussed in person with @stefan-kracht .

I decided the name should start with "Tooltip" after all, because it requires to be wrapped in a TooltipProvider.

We discussed making the tooltip optional in the future, but my current thoughts on this are: It might make things complicated (regarding the requirement of having a parent TooltipProvider), and my main reason to componentize this was the boilerplate code required for the tooltip. For use cases where we just need a badge without tooltip, I am not sure using a component (other than `design/Badge` is justified.

I have added the component to the instances list and updated the e2e tests accordingly (because test-ids have changed).

### DIFF
--- a/e2e/instances/list/index.spec.ts
+++ b/e2e/instances/list/index.spec.ts
@@ -115,7 +115,7 @@ test("it renders the instance item correctly for failed and success status", asy
       `instance-row-wrap-${instance.instance}`
     );
     const instanceItemId = page.getByTestId(
-      `instance-row-id-${instance.instance}`
+      `tooltip-copy-badge-${instance.instance}`
     );
     await expect(
       instanceItemId,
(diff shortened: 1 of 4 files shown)


stefan-kracht (approved):
I think it's a very good idea to have this as a separate component. As we just discussed, I made a slide suggestion on the API of this component. Since the component looks very much like the Badge Component, just with additional features on top of it, it would be nice if it could be used in a very similar way.

So the following code:

`<TooltipCopyBadge value={instance.id} displayValue={instance.id.slice(0, 8)}/>`

would change to 

`<TooltipCopyBadge value={instance.id} variant="outline">{instance.id.slice(0, 8)}</TooltipCopyBadge>`

I also added some stories to outline the flexibility

CleanShot 2023-08-08 at 13 33 48

PR #442, opened 2023-08-08
[DIR-196] Events page

sebxian:
Preview: https://direktiv-ui-refs-pull-442-merge.direktiv.dev/

## To do:

My list of finishing touches left here FYI. Feel free to skip to "Notes for Review".

I postponed some of these (unchecked) and created issues. Overlaps with "notes for review" where I explain relevant stuff.

* [x] click on event row: Open event content in dialog with editor, button to "retrigger" event. 
* [x] click on listener "target": Link to workflow or instance depending on event listener type.
* [x] Header with button over events table: Margins need to be fine-tuned between history and listeners.
* Postponed - DIR-737 ~~Truncate events list when new events come in? I think currently, a new event being added may cause the list to grow over the limit.~~
* [x] Width of events detail dialog?
* ~~Error handling when sending events?~~ see DIR-733
* [x] Make sure form labels and testids are correct
* [x] Ticket for e2e tests - DIR-741
* [x] ~~skipped for now -- [Antenna](https://lucide.dev/icons/antenna) would be a nice event-listener icon. This requires updating Lucide, which breaks some Types that will need to be fixed.~~ [EDIT from @stefan-kracht]: updated the icon in 225da73a5d2453455aef969e56668b776077a0c3 (I also had to update some types to make TS happy c3cbc35bd501010731e37e35b6b9e7d1b050513f)

## Notes for review

I am marking this as "ready for review" in a moment, but it feels a bit "hot", some things may be less polished than I'd like and I haven't been able to thoroughly double-check everything. But it's probably good to start the review process nevertheless. @stefan-kracht, feel free to suggest or make improvements.

I would argue some of the following are due to back-end bugs that should be fixed there. But that needs to be discussed with the team. I have created issues in linear (linked below). If you agree, we should bring these up with the team on Monday. I've also briefly mentioned them to Alexander and Alan in Slack.

* **Scope of testing**: I admit I haven't found the time to extensively test the event-listeners page with lots of events regarding pagination! I don't expect problems as the mechanics are mostly identical to those on the events page.
* **Tab navigation**: I don't use the router for navigating between the tabs as we discussed. I think it would be worth adding that if it's not much work, I found it a bit annoying that the browser didn't remember where I was while testing and navigating between pages.
* **Event listener types**: [This workflow](https://direktiv-ui-refs-pull-442-merge.direktiv.dev/sebxian/explorer/workflow/active/consume.yaml), when started as an instance, will be listed as an event listener for type "resume instance". It fails after a while (due to some timeout I did not investigate further), so you'll have to start a new instance manually to test it.
* **Event listener links to target**: In the "target" column, there will be a link to the instance or workflow depending on the listener type. For workflows, it only works with root-level files at the moment, because the API doesn't return the full paths for workflows. I have created an issue DIR-740. If it isn't fixed in the back-end, we need to fix it in the front-end or remove the linking feature before release.
* **Viewing events**: You can click a row and it will open a dialog with the editor showing the "cloudevent" part of the event. There are two issues with this currently that are probably back-end bugs (according to Alexander, an issue DIR-739 has been created to track this):
  * the data isn't in JSON format, as it should be (the Editor currently is set to format its content as JSON, hoping this will be fixed soon).
  * the "cloudevent" part is empty on brand new events (that have come in via streaming), in these cases the editor stays empty (again, assuming that is a back-end bug that will be fixed).
* **Retriggering events**: When retriggering and event that has an event listener, you will see the triggered target executed in the "instances" page (you have to navigate there yourself. When sending the event, we don't know if there's a listener). But the event itself will not re-appear on top of the list. I assume it's timestamp isn't updated and it isn't emitted as a new event in the stream.
* **Sending new events** When sending new events, we just validate the event body as a string. I added a ticket [DIR-733] to revisit this later and maybe improve validation (or we should just rely on back-end validation and show the errors from the back-end response).
* **Pagination** Issue DIR-737 created to track these issues related to: Updating pagination (and invalidate cache) whenever new events come in.
* **Time tooltip** I noticed the Tooltip over a relative time (formatted with moment) shows the "raw" time in all lists that I am aware of. Should we generally change this to show a more human-readable time? (that's something we could leave for post 0.8.0, however)


src/pages/namespace/Events/History/EventsList.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const noResults = isFetched && data?.events.results.length === 0;
+  const showPagination = numberOfResults > itemsPerPage;
+  const hasFilters = !!Object.keys(filters).length;
+
+  return (
+    <div className="flex grow flex-col gap-y-3 p-5">
+      <div className="flex flex-row justify-between">
+        <h3 className="flex items-center gap-x-2 pb-2 pt-1 font-bold">
+          <Radio className="h-5" />
+          {t("pages.events.history.title")} {eventDialog ? "open" : "close"}
```
stefan-kracht, line 66: I this `{eventDialog ? "open" : "close"}` some old debug code you forgot to remove? 

I was thinking if we should remove the headlines because it kind of only repeats what the tab is showing. What do you think? The "send new event" button could go on the opposite site of the filter button
CleanShot 2023-08-11 at 17 57 40@2x

src/pages/namespace/Events/Listeners/NoResult.tsx
```diff
@@ -0,0 +1,13 @@
+import { Calendar } from "lucide-react";
+
+const NoResult = ({ message }: { message: string }) => (
+  <div
+    className="flex flex-col items-center justify-center gap-1 p-10"
+    data-testid="listener-no-result"
+  >
+    <Calendar />
```
stefan-kracht, line 8: I think you forgot to update the Calendar icon. When you search for `<Calendar />` in the `./src/pages/namespace/Events`folder, I find 4 currencies that might need to be updated.

stefan-kracht (changes requested):
This is nice work! I'm super happy that the filters are already in scope of this PR. 

I found it a bit hard to get my head around how events work and how they can be triggered, but I think I get the general idea now.

---

> Time tooltip I noticed the Tooltip over a relative time (formatted with moment) shows the "raw" time in all lists that I am aware of. Should we generally change this to show a more human-readable time? (that's something we could leave for post 0.8.0, however)

I was asking myself the same thing the other day, and I think we should. My initial argument for not doing this was to show the user the raw value because it might be useful for debugging (like digging in some logs in grafana) but I don't know if that makes any sense at all 😄 

---

## I was able to fix some of your open tasks:

> [Antenna](https://lucide.dev/icons/antenna) would be a nice event-listener icon. This requires updating Lucide, which breaks some Types that will need to be fixed.

I [updated the package](https://github.com/direktiv/direktiv-ui/commit/225da73a5d2453455aef969e56668b776077a0c3) and you were right. There were some issues we had to make for TS to be happy after the update. They seem to have changed some TS definitions in `v0.260.0`. We use `React.FC<React.SVGProps<SVGSVGElement>>` at some places to describe an icon as a type. The Icons don't fulfill that type anymore, and we would have to change it to `ForwardRefExoticComponent<RefAttributes<SVGSVGElement>>` to make it work. However, I decided against that unreadable generic type and used the `LucideIcon` type instead. It now depends on the library and will not allow other SVG icons (they add two props that are not part of an svg element: `size` and `absoluteStrokeWidth`), but since we are using that library for our icons anyway, I found this to be the better and intuitive solution. Also, it's easy to move away from that type at any time.

>  I don't use the router for navigating between the tabs as we discussed. I think it would be worth adding that if it's not much work, I found it a bit annoying that the browser didn't remember where I was while testing and navigating between pages.

[I did this as well](https://github.com/direktiv/direktiv-ui/pull/442/commits/c0c1f33d7056a860d06f47e7e7301db018ea28db). There is a tiny issue that we already have on the explorer page, but will be solved in a [different ticked](https://linear.app/direktiv/issue/DIR-482/explorer-must-be-marked-as-active-menu-item-when-on-workflow-page) (I made a note in the comments to fix it on the events page as well).

I also added the following: 
When we can not run a workflow via the UI, the user should see a message. I added a toast that shows any error message that the server might send back:
CleanShot 2023-08-11 at 17 34 01@2x

src/pages/namespace/Events/History/EventsList.tsx
```diff
@@ -0,0 +1,145 @@
...
+        </h3>
+
+        <SendEvent />
+      </div>
+
+      {/* <div className="flex flex-row justify-between">
+        <h3 className="flex items-center gap-x-2 pt-1 font-bold"></h3>
+      </div> */}
```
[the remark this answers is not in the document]
sebxian: Fixed in 6eb9f631c8a686c25b79b05961354646677cfb90

src/api/eventListeners/query/get.ts
```diff
@@ -0,0 +1,151 @@
...
+
+const updateCache = (
+  oldData: EventListenersListSchemaType | undefined,
+  message: EventListenersListSchemaType
+) => {
+  if (!oldData) {
+    return message;
+  }
+  /**
+   * TODO: copied from events. Does this behave the same?
```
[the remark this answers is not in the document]
sebxian: Fixed in 6eb9f631c8a686c25b79b05961354646677cfb90

You're right, removed TODO and updated "logs" to "results". I copied the comment over with that method and thought it can't hurt to keep it.

src/pages/namespace/Events/History/EventsList.tsx
sebxian: Very good catch, thanks. Debug output removed in 6eb9f631c8a686c25b79b05961354646677cfb90.

I agree about the duplicate headline and will attempt a fix in a separate commit.

sebxian: I removed the redundant headers as suggested in 0d4483357f7211fea8e40f82516a8eb7884866f7.

It can get a bit crowded in the filters area (but that is already the case even without the "send event" button). I marked the send-event into a primary button to make it stand out more. What do you think?

Screenshot from 2023-08-14 17-37-28
Screenshot from 2023-08-14 17-37-16

src/api/events/mutate/sendEvent.ts
```diff
@@ -0,0 +1,66 @@
...
+        "content-type": "application/cloudevents+json",
+      },
+      urlParams: {
+        namespace,
+      },
+    });
+
+  return useMutation({
+    mutationFn,
+    // no cache invalidation because events are updated via streaming
```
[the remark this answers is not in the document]
sebxian: As discussed in person, I briefly tried out adding cache invalidation and decided against it because (a) we don't need it in the current use case (due to streaming) and (b) it is not trivial to implement (so the time investment isn't justified right now).

I updated the comment regarding this in efd8d9523b04af1a7b6d906d634bb65a3ed23c75

sebxian:
@stefan-kracht, thank you so much for the changes to this branch, that was very helpful :+1: 

And also for your valuable feedback.

I went through all your comments and the changes are ready for review again.

I also simplified updating the cache for paginated streams in 6b9ad58ed73d5f48b305478becc2b604533f338f. I found out the stream always returns a whole page of results on every update, so it is not necessary to merge old results with the update in this case. This also solves the small bug that there could be more than 10 results on a page when new events came in via streaming.

src/pages/namespace/Events/History/EventsList.tsx
stefan-kracht: Using a primary button is a very good idea. It now differs very well from the filter buttons

stefan-kracht:
Looks very nice. Feel free to merge!

stefan-kracht (approved):

PR #443, opened 2023-08-10
[DIR-626] add pagination and search to settings page

stefan-kracht:
**[Preview 🚀](https://direktiv-ui-refs-pull-443-merge.direktiv.dev/)**

This PR adds a client side search to secrets, registries and variables CleanShot 2023-08-10 at 15 08 02@2x

- Made header of secrets, registries and variables responsive
- I changed every "+" Button to be like "+ Create Variable". It looks better on mobile, because the button is much bigger. And I even like it better to very explicit, and it also matches "+ send new event" button on the new events page CleanShot 2023-08-10 at 14 25 04@2x
- Changed modal button from primary variant to default variant. I noticed that we never use a primary button in a modal elsewhere. I'm open to change that in the future but I remove the primary variant for now, just for consistency

## Follow up tickets
- [E2E tests](https://linear.app/direktiv/issue/DIR-738/pagination-and-filter-on-the-settingspage)

PR #445, opened 2023-08-11
[DIR-622] test registry credentials

stefan-kracht:
This PR is an add-on to [this PR](https://github.com/direktiv/direktiv-ui/pull/443) to avoid merge conflicts. It implements the Test Connection Button on the registry page.

**[Preview 🚀](https://direktiv-ui-refs-pull-445-merge.direktiv.dev/)**

https://github.com/direktiv/direktiv-ui/assets/121789579/c04592d4-833d-4faf-914c-2153bd40b9ba

While developing this feature, I found out that the API behind this button is more a validation and not a real connection test (e.g. any url starting with https://hub.docker.com passes the connection test, username and password does not seem to matter). We should discuss, whether we should just mimic this behavior on the client and make this a real validation and not implying to check the credentials and the actual existent of the repo.

## Follow up tickets
- [E2E-Tests](https://linear.app/direktiv/issue/DIR-742/test-connection-button-when-adding-a-registry)

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -329,7 +329,7 @@
           "user": "User",
           "password": "Password",
           "testConnectionBtn": {
-            "label": "Test Connection",
+            "label": "Test connection",
             "success": "Connection successful",
             "error": "Connection failed"
           }
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -327,7 +327,8 @@
           "description": "Add a container registry",
           "url": "URL",
           "user": "User",
-          "password": "Password"
+          "password": "Password",
+          "testConnectionBtn": "Test Connection"
         }
       },
       "variables": {
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -328,7 +328,11 @@
           "url": "URL",
           "user": "User",
           "password": "Password",
-          "testConnectionBtn": "Test Connection"
+          "testConnectionBtn": {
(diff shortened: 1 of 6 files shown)


public/locales/en/translation.json
```diff
@@ -327,7 +327,12 @@
           "description": "Add a container registry",
           "url": "URL",
           "user": "User",
-          "password": "Password"
+          "password": "Password",
+          "testConnectionBtn": {
+            "label": "Test Connection",
```
sebxian, line 332:
```suggestion
            "label": "Test connection",
```

sebxian (commented):
I really like the implementation, and the look & feel of the button :+1: 

But regarding your comment and how the back-end behaves, I would actually vote against merging this and rely front-end validation instead. Suggesting that login credentials are correct when we don't really know could cause more harm than good in terms of UX and customer support.

We already have this to validate the input data:

```
export const RegistryFormSchema = z.object({
  url: z.string().url(),
  user: z.string().nonempty(),
  password: z.string().nonempty(),
});
```

Which is not much, but I am not sure if the current connection testing adds any value on top of that. I mean, it seems to check if the url points to `hub.docker.com`, but does that restriction really make sense? (If so, it would be easy to validate for this client-side).

stefan-kracht:
Sorry, I accidentally merged :D But [I added this follow up ticket](https://linear.app/direktiv/issue/DIR-748/test-connection-button-should-be-client-side-validation) an will take care of it

PR #447, opened 2023-08-16
[DIR-208] workflow settings

sebxian:
## Description

[Preview](https://direktiv-ui-refs-pull-447-merge.direktiv.dev/)

This only adds workflow-variables to the workflow-settings page, because the "Log to Event" and "Attributes" features in the old UI are unused or broken and it was decided not to implement them in the new UI.

We could even think about renaming the tab to "variables", but I left it "settings" for now which leaves room for future additions.

### NoResults Component

I created a generic "NoResults" component because we were re-using the same code a lot. I also added examples under "Table" in Storybook, and compared all occurrences visually to make sure [they look](https://direktiv-ui-refs-heads-feature-redesign.direktiv.dev/test/monitoring) the same [as before](https://direktiv-ui-refs-heads-feature-redesign.direktiv.dev/test/monitoring) (yay for preview links :rocket:).

### Errors when fetching variables

I added an error Alert to the Variable Edit dialog (see screenshot). There are two drawbacks of this solution, but those are not easy to solve so I think this is good enough for now:
* one could argue a toast should be shown and no dialog should be rendered at all when there is an error
* there is flickering because the full form is rendered before we know there's an error

But there are also reasons to stick with this approach:
* Opening the dialog after we know fetching was successful would make the UI feel sluggish
* The error should never occur, because we would not allow the user to open the dialog for an element that doesn't exist. It only occurs right now because of a backend bug: workflow-variables are listed on the namespace settings page, where they don't belong, and thus they can't be fetched for editing. I expect that to be fixed in the backend (DIR-754).

image

### Points for discussion

* Most of the codes for viewing and editing variables was recycled from the settings page, in some cases it was even possible to import components without changes. In the other cases, there is room for future refactoring to reduce code duplication (-> DIR-753). I decided not to duplicate identical translation strings to the `workflow.settings.variables` namespace because I think it would be better to consolidate the forms in the first place. So `workflow.settings.variables` only has values that differ.
* 653736e4c741c545cd6eb4632aeda06982a80ef0 feels a bit like a workaround: I passed `{ path }` down into components so I would be able to escape for the case that `path === undefined`. It doesn't make sense as an optional property, because we always need the path for all queries related to workflow variables. It should actually never be undefined on a workflow route, but when importing it as `const { path } = pages.explorer.useParams();`, that method isn't smart enough to know that (and maybe it shouldn't?).
* I updated `nginx.conf` in 76d4f5af5544a8badc8f43ed478c726860b2deb0 to fix the preview links, but later I learned that this value is overwritten in the CI pipeline. UPDATE: Have edited nginx.conf again in bd9adf32edd18198960b657996874dc7920c74b5 with a placeholder and comment.


conf/nginx.conf
```diff
@@ -9,7 +9,7 @@ server {
   }
 
   location ^~/api/ {
-          proxy_pass https://135.181.25.123/api/;
+          proxy_pass https://172.232.37.46/api/;
```
sebxian, line 12:
```suggestion
          proxy_pass https://127.0.0.1/api/; will be replaced according to CI pipeline config
```

src/design/Table/index.tsx
```diff
@@ -108,3 +109,19 @@ export const TableRow = React.forwardRef<
...
+}: {
+  message: string;
+  icon?: LucideIcon;
+}) => (
+  <div
+    className="flex grow flex-col items-center justify-center gap-1 p-10"
+    data-testid="no-result"
+  >
+    {Icon && <Icon />}
+    <span className="text-center text-sm">{message}</span>
```
stefan-kracht, line 125: Could the `message` prop just be `children` here? I'm a big fan of using children whenever it makes sense` <Component>text</Component>` looks a bit more react-ish in comparison to `<Component text="text" />`

stefan-kracht (changes requested):
Very nice work @sebxian. The only improvements that I might have had are already covered in DIR-753. Plus one comment in the code....

---

> We could even think about renaming the tab to "variables", but I left it "settings" for now which leaves room for future additions.

I think "settings" is fine. I'm sure it will expand at some point anyways.

---

> I created a generic "NoResults" component because we were re-using the same code a lot. 

That's a very good one. In the future, we might expand it to cover `src/pages/namespace/Explorer/Tree/NoResult.tsx` as well. This one also has a button. The services will have a similar one.

CleanShot 2023-08-20 at 18 21 30@2x

---

About Errors when fetching variables: I like how you solved it! 👌

--- 

Could we make a very simple E2E tests ticket for Aaron? It would be very easy to give it to him because the tests should be the same as in the namespace settings. I would naturally  also write a test for the bug were workflow variables were leaking to the namespace settings even if this should technically be a part of the backend E2E tests. What do you think about it?

---

> It should actually never be undefined on a workflow route, but when importing it as const { path } = pages.explorer.useParams();, that method isn't smart enough to know that (and maybe it shouldn't?).

That is actually expected behavior. e.g.  pages.explorer.useParams() is meant to be callable on any side and it will return undefined when you are not on an explorer page. This is very handy for components like the breadcrumb (it is used on every page and has to be reactive to the page context). We could introduce something like pages.explorer.useParamsOrThrow() that never returns undefined, but will throw an error when you use it on the wrong page. 

Generally, my hopes for [TanStack Router](https://tanstack.com/router/v1) are very high. It is still in beta but offers a lot of solutions to react router typescript problems. It should not be too hard to migrate and should make the `src/util/router/pages.tsx` file much more readable (and even more type safe). So we can keep this solution for now and I'm sure it will be a bit more elegant in the future.

src/design/Table/index.tsx
sebxian: Good point! I applied this in the last 2 commits.

sebxian:
> That's a very good one. In the future, we might expand it to cover src/pages/namespace/Explorer/Tree/NoResult.tsx as well. This one also has a button. The services will have a similar one.

I deliberately left it our for now because I preferred to keep the component as simple as possible and not over-engineer it with additions that have only a single use-case. But I'm not against re-evaluating this when we get more similar use cases.

Regarding the e2e ticket, I created it as DIR-755. (I also created an e2e ticket for the toggle active/inactive button under DIR-756).

So this should be ready for review again, @stefan-kracht

stefan-kracht (approved):

PR #448, opened 2023-08-17
[DIR-209] set live status on workflow

sebxian:
[Preview](https://direktiv-ui-refs-pull-448-merge.direktiv.dev/)

This adds the ability to turn the workflow active / inactive.

It's implemented on top of the branch for the workflow settings page in PR #447.

I made the dropdown trigger variant="primary" as well, but that doesn't look good when the main button is inactive:

image

So I changed the dropdown trigger to "outline", but only when the other button is inactive. It's not 100% logical to have the dropdown button active in the second example, but that way it blends in naturally with the main button.

image
image

* [x] should we add a tooltip to the button when the workflow is inactive? I think the button is already pretty self-explanatory (at least when you investigate and click on the dropdown)

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -172,10 +172,10 @@
         },
         "headline": "Latest",
         "updated": "Updated {{relativeTime}} ago",
-        "runBtn": {
-          "runWorkflow": "Run",
-          "setActive": "Set active",
-          "setInactive": "Set inactive"
+        "runBtn": "Run",
+        "toggleActiveBtn": {
+          "setActive": "Activate workflow",
+          "setInactive": "Deactivate workflow"
         },
         "editor": {
           "unsavedNote": "unsaved changes",
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -169,7 +169,11 @@
         },
         "headline": "Latest",
         "updated": "Updated {{relativeTime}} ago",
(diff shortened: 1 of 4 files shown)


stefan-kracht:
I know it's still a draft, but I already went through the code and reviewed it. Looks very good to me!

> should we add a tooltip to the button when the workflow is inactive? I think the button is already pretty self-explanatory (at least when you investigate and click on the dropdown

I think we don't need a tooltip here

> So I changed the dropdown trigger to "outline", but only when the other button is inactive. It's not 100% logical to have the dropdown button active in the second example, but that way it blends in naturally with the main button.

I'm not 100 % sure about the design yet. Mostly because the ButtonBar was not optimized for a primary button (In dark mode the primary button gets a border, which it never has, and it looks kind of weird) and the inactive button next to an active button somehow just looks broken :D  What do you think about solving it pragmatically with just two buttons (the icon one would need a tooltip, though).

CleanShot 2023-08-20 at 19 08 53@2x

sebxian:
@stefan-kracht thanks for the feedback, I'd actually forgotten to mark it as ready for review.

I've implemented the suggested change with the separate button. I agree it looks cleaner (the code feels cleaner too).

I've marked it as ready for review now.

stefan-kracht (approved):
Looks great, merge it whenever you like 🚀

PR #450, opened 2023-08-17
Add selfhosted runner

h8718:
## Description

^


stefan-kracht:
Can you give me a very short info about this? This changes the playwright action to run on a self hosted machine, right?

h8718:
Yes, it will use a selfhosted Machine instead of the default one. If it
turnout to be slower we may change it back or add an additional self hosted
Machine.

Stefan ***@***.***> schrieb am Fr., 18. Aug. 2023, 07:42:

PR #451, opened 2023-08-18
[DIR-734] Toast too big

porleaaron:
### What's done

1.  Remove padding and margin around Toast 

### Question

1. Do we have use case of "toasts list", many toasts on one screen ?


stefan-kracht:
> Do we have use case of "toasts list", many toasts on one screen ?

I'm no really sure what you mean. How would that look like? 

Otherwise, it looks good! I just noticed an issue on very small screens that is unreleated to this ticket but I think it would make sense to fix this right away. There should be more padding on the right: 

CleanShot 2023-08-18 at 07 29 36@2x

PR #458, opened 2023-08-28
[DIR-482] active menu bug

stefan-kracht:
Fixes issue where not all menu items were shows as active, when the page has multiple subpages.

Previous approach relied on the [isActive](https://reactrouter.com/en/main/components/nav-link#classname) parameter of the  [NavLink-Component](https://reactrouter.com/en/main/components/nav-link). But it simply does not seem to support subpages.

Fixed by adding a naming convention to our pages config (`src/util/router/pages.tsx`) like for example

```typescript
someNavItem: {
  // ...
  route: {
    // ...
    handle: { someNavItem: true }
    // ...
  }
}
``` 

I was not able to make it type safe for now, so we have to remember to that for now.


src/util/router/pages.tsx
```diff
@@ -216,6 +216,7 @@ export const pages: PageType = {
     },
     route: {
       path: "explorer/",
+      handle: { explorer: true },
```
sebxian, line 219: What do these properties in "handle" represent exactly? I wonder if the name (`explorer` in this case) could be changed to something containing more semantic information.

stefan-kracht: The name explorer in this case must exactly match the key of the page, so that the Navigation component has one programmatic way to detect the active (root) menu

`const isActive = checkHandlerInMatcher(secondPathLevel, key);`

sebxian (approved):

PR #460, opened 2023-08-29
[DIR-756] e2e - toggle workflow active inactive

porleaaron:
### What's done

1.  e2e test for workflow active/inactive toggle

### What needs to be discussed

1.  Is this enough to open the Run Workflow modal to check if the "run" button works after reactivate?


e2e/explorer/index.spec.ts
```diff
@@ -407,3 +409,51 @@ test(`it is possible to rename a directory`, async ({ page }) => {
...
+    iconPowerOn,
+    "next state should be inactive, shows the power-on button"
+  ).toBeVisible();
+  await expect(
+    btnRun,
+    "next state should be inactive, run button is inactive"
+  ).toBeDisabled();
+
+  await btnRun.click({ force: true }); //nothing happens on this click
+  await expect(btnRun, "run button is still inactive").toBeDisabled();
```
sebxian, line 449: If possible, it would be good to check the "run-workflow-dialog" isn't opened here.

This is a bit tricky to implement because you need to make sure it waits for a bit to avoid a false test result (if you check immediately after clicking, the modal will probably always be invisibe even if it is opening).

So to make sure your assertion works, you have to put it in another place where you know that the modal will open, and make sure that it will fail.

sebxian (changes requested):
Thanks, good work, @porleaaron !

To answer your question, yes, it's enough to see that the modal is opened at the end.

I have two requests for changes:
1. see comment in the code
2. Please move this test to the file `e2e/explorer/workflow/active.spec.ts`.

e2e/explorer/index.spec.ts
porleaaron: I have no way to know that nothing happens on page after clicking the run button which is disabled
I think we can simply wait for 1 sec and check the dialog is opened

sebxian: This was implemented with waiting for 1 sec.

sebxian (approved):
I updated the PR to address some small remaining issues, it's ready to merge now.

PR #454, opened 2023-08-25
[DIR-250] add authentication modal

stefan-kracht:
This PR adds the authentication modal when needed

[Preview 🚀](https://direktiv-ui-refs-pull-454-merge.direktiv.dev/)

https://github.com/direktiv/direktiv-ui/assets/121789579/5f30a8ba-c833-4c14-93e6-e8892a4443e6

The [E2E ticket](https://linear.app/direktiv/issue/DIR-772/login-page) is a good starting point to get an idea of the scope of everything.

## Approach

I added a component as a top level wrapper with an early return, which makes it very simple in terms of url handling. The url is never touched (e.g. when I open a bookmarked workflow and have no auth key in my localstorage anymore, I can enter the key and will get to the desired workflow afterwards).

## Notes
- I also removed an old `VITE_APP_VERSION` env variable that is not needed anymore
- I completely removed our old component for setting the api key since this was just a workaround/placeholder
- I made some small changes to the file structure of the versions endpoint (it was one of the first endpoints I did and the structure of the files was not consistend with all other api folders)
- I cleaned up the user menu (images below)
  -  I removed the username (because there is none)
  - when auth is required, it shows a usericon instead of the first characters of the username
  - when auth is not required, it shows a settings icon and will not show a logout button
  - Updated the help section
    - added a link to the docs
    - removed jqPlayground (Jens is using this a lot it will probably be a main menu, but can be discussed further when we implement it )
    - removed "show API commands" (This belongs somewhere in the explorer, because it is just useable there, but can be discussed further when we implement it)

CleanShot 2023-08-25 at 17 18 03@2x
CleanShot 2023-08-25 at 17 17 47@2x

## Follow up tickets
- [E2E Test for login modal](https://linear.app/direktiv/issue/DIR-772/login-page)

--- 
EDIT: I just found out that the logo did not work very well in light mode. So, I adjusted the modal API a bit to allow us to edit the overlay appearance. I like this a lot better now, it now looks more like Card and not an overlay (technically it does not lay above anything, so the gray bg loos weird)

Before:

CleanShot 2023-08-25 at 18 30 31@2x

After: 
CleanShot 2023-08-25 at 18 38 38@2x

PR #461, opened 2023-08-29
[DIR-555] API commands button

stefan-kracht:
This PR implements the API commands button. I adjusted the available commands and made them interactive.

### [Preview 🚀](https://direktiv-ui-refs-pull-461-merge.direktiv.dev/)

CleanShot 2023-08-29 at 18 53 26

https://github.com/direktiv/direktiv-ui/assets/121789579/e6c5bd63-ce9c-4c33-81cc-e8c938c6dc19

## Note
- Please don't review before merging https://github.com/direktiv/direktiv-ui/pull/454. This has the branch `stefankracht/dir-250-add-authentication` already merged, because I needed an easy method to access the auth status and token
- I have a `useEffect` that bugs me, but I can not get rid of easily. I think it is good for now
- I stumbled upon a bug in our Editor and needed to move the place where `onChange` was called. The old behavior passed a reference of this function when it mounts, and it will never get updated. (a650930f8c3b9f2b21ee864357b142fe6354e3df)   Example: 
```typescript
const [var, setVar] = useState(1)
    // ...
   <Editor
     // ...
      onChange={(data) => {
         console.log(var) // this was always 1, even when the state was update somewhere else via setVar
      }}
      theme={theme ?? undefined}
    />
``` 
After fixing this, Playwright caught an issue: the problem was on the workflow edit page. When the user restored, an old revision, it accidentally showed "unsaved changes" at the bottom. This was caused by an useEffect. 

## Follow-up tickets
- [E2E test](https://linear.app/direktiv/issue/DIR-782/test-api-commands-button)
- [There is another API Commands button that I was not aware of, but it's okay to do is after the release](https://linear.app/direktiv/issue/DIR-783/add-api-commands-button-to-the-file-browser-as-well)


sebxian (commented):
Created ticket for later code review (DIR-799). Ready to merge for further testing in `feature/redesign`.

It might be best to merge this after [454](https://github.com/direktiv/direktiv-ui/pull/454/commits). I made some spelling fixes there that may conflict with this branch. If you have to resolve conflicts, make sure the changes in translation.json and e2e/usermenu.spec.ts are consistent (e.g., /Switch to Dark mode/).

PR #465, opened 2023-09-01
[DIR-755] e2e - workflow variables

porleaaron:
### What's done

1. Prove it's possible to navigate to settings tab & pagination works
2. Prove it's possible to create/update/delete variables

###  Questions 

1.  Currently there's no validation for mimeTypes


sebxian:
I have made extensive changes to this PR to align it with our best practices for writing e2e tests.

Hence, it would be good if you could review the end version as well, @stefan-kracht.

stefan-kracht (approved):
Looks very good to me! At some point, we can add a test to make sure file upload works as well, but I'm pretty sure this feature did not exist when the task for this ticket was created.

EDIT:  After writing this, I found [a ticket](https://linear.app/direktiv/issue/DIR-807/file-uploads) for that in our e2e backlog :)

PR #466, opened 2023-09-03
[DIR-205] workflow overview page

sebxian:
[Preview](https://direktiv-ui-refs-pull-466-merge.direktiv.dev/)

## Layout Notes

Fine-tuned the layout of the metrics / traffic distribution graphics to take as much space as 10 rows of results in the instances table (so there will be empty space with fewer instances): 
image
Screenshot from 2023-09-06 13-46-04

Added a "Services" tab because it would be too complicated to navigate to service-revisions and revision-details (with pod logs) within the overview tab. When on the services tab, without additional query params, it lists all wf-services, additional queryparams are added to view revisions of one wf-service and details (with pod logs) of one revision.

I re-used the service components (and updated them to handle both use cases). This is how they now look in the worfklow-services context. Their layout and behavior for namespace services shouldn't have changed (but it is worth double checking during review!)

Screenshot from 2023-09-14 15-54-48
Screenshot from 2023-09-14 15-54-55
Screenshot from 2023-09-14 16-03-06
Screenshot from 2023-09-14 16-03-47

## Notable changes to existing code
* I re-used the components and api methods originally created for namespace-services. In general, two things were necessary to adapt them to work with workflow-services:
  * In the api methods, I passed in the workflow path and the service version property as optional parameters. When both exist (they are both needed), a different URL specific for workflow-services is used in the request. This applies both to streaming and non-streaming endpoints. 
  * I couldn't re-use the existing services pages in full (for example, the header layout etc. are different in the workflow context), so I extracted some of their code to smaller re-usable components that can be imported regardless of context. 

## Follow up issues
* e2e tests: DIR-845
* [tremor.so](http://tremor.so/) is used on the workflow overview page now. We should re-evaluate what framework to use for data visualization. Tremor turned out to be quite limited in terms of (a) customizing the colors used in the charts and (b) it seems it relies a bit too much on being used inside of tremor's own parent <Card> components instead of our own for the purpose of calculating the sizes of the SVGs. Probably b can be solved (I was able to make it work but didn't have enough time to investigate more deeply), but I think (a) is a big enough problem to consider altermatives. I think flowbite is worth a closer look. -> DIR-846
* DIR-847: Tweaks to the layout and navigation to consider for the future.


stefan-kracht:
As discussed [I will review this after the launch](https://linear.app/direktiv/issue/DIR-848/workflow-overview).

PR #468, opened 2023-09-04
[DIR-552] add 'insecure' option to mirrors

sebxian:
[Preview link](https://direktiv-ui-refs-pull-468-merge.direktiv.dev/insecure3/mirror/) 

This adds a checkbox for the "insecure" property to the mirror create/edit form.

Currently, there seems to be a backend bug documented in the comments of [DIR-552] that doesn't let you change the "insecure" property from true to false. But the frontend implementation should work as intended (i.e., it sends the appropriate requests).

image


sebxian:
@stefan-kracht , I think this is ready to review, or merge and review later if that's ok with you.

I updated the wording after feedback in the office chat & merged the current feature-redesign into this. 

There is still a backend bug described in the comments of DIR-552, but that shouldn't stop us.

stefan-kracht:
Great, I added my [postponed review](https://linear.app/direktiv/issue/DIR-818/review-add-insecure-option-to-mirrors) and will merge right away.

PR #472, opened 2023-09-06
[DIR-713] make secrets editable

stefan-kracht:
**NOTE: https://github.com/direktiv/direktiv-ui/pull/463 is already merged in, please review it first**

**[Preview 🚀](https://direktiv-ui-refs-pull-472-merge.direktiv.dev/)**

https://github.com/direktiv/direktiv-ui/assets/121789579/563dda80-4feb-4bf1-818c-806690546118

This PR makes secrets editable and highlights uninitialized secrets.

I also removed the manual cache update after creating/editing and just use cache invalidation for the following reasons:
- cache update mistakenly sorted results, this does not match the server result (anymore?)
- it is possible to enter an existing variable name in the Create Variable dialog. This actually updates the old one, the cache update mechanism will show the same entry twice until the user reloads or useQuery refetches
- both points above illustrate, why it is probably a good idea to prefer cache invalidation in the future. It also makes the code much more readable.

### Follow-up tickets
-  [E2E-test](https://linear.app/direktiv/issue/DIR-808/update-secrets-and-test-initialized-attribute)

### DIFF
--- a/e2e/settings/index.spec.ts
+++ b/e2e/settings/index.spec.ts
@@ -2,7 +2,7 @@ import { createNamespace, deleteNamespace } from "../utils/namespace";
 import { expect, test } from "@playwright/test";
 
 import { BroadcastsSchemaKeys } from "~/api/broadcasts/schema";
-import { TextMimeTypeSchema } from "~/pages/namespace/Settings/Variables/MimeTypeSelect";
+import { EditorMimeTypeSchema } from "~/pages/namespace/Settings/Variables/MimeTypeSelect";
 import { actionWaitForSuccessToast } from "../explorer/workflow/utils";
 import { createBroadcasts } from "../utils/broadcasts";
 import { createRegistries } from "../utils/registries";
@@ -11,7 +11,7 @@ import { createVariables } from "../utils/variables";
 import { faker } from "@faker-js/faker";
 import { radixClick } from "../utils/testutils";
 
-const { options } = TextMimeTypeSchema;
+const { options } = EditorMimeTypeSchema;
 
 let namespace = "";
 
--- a/e2e/utils/secrets.ts
+++ b/e2e/utils/secrets.ts
@@ -1,6 +1,6 @@
-import { createSecret } from "~/api/secrets/mutate/updateSecret";
 import { faker } from "@faker-js/faker";
 import { headers } from "./testutils";
+import { updateSecret } from "~/api/secrets/mutate/updateSecret";
 
 export const createSecrets = async (namespace: string, amount = 5) => {
   const secrets = Array.from({ length: amount }, () => ({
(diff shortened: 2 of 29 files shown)


sebxian (changes requested):
Nice work! 

I merged in feature/redesign and pushed a commit renaming variables (the same changes we had made in a previously merged branch, but git overwrote them with this branch's version).

Unfortunately I found some bugs with the file-upload form while testing. 

Since we already merged the branch with these changes, but there was some overlap with this branch, I note them here. I think we should try to fix them before release.

First, I can upload a file in the secrets form and a binary (png) will be treated as text and loaded into the editor:
Screenshot from 2023-09-14 10-00-14
When I try to save that, it (expectedly) returns a backend error.

In the variables form (including workflow variables), the behavior is correct.

UPDATE: Fixed (push pending): I cannot download a binary namespace-variable due to a zod schema error:
Screenshot from 2023-09-14 10-11-03

The request performed when clicking "Download" is correct and it does return the file when I test it in Postman (`http://localhost:3000/api/namespaces/sebxian/vars/png-image`). But the server response doesn't have a "content-type" header.

I also checked the request we send when uploading the file and it does have the "content-type" header.

Interestingly, the file download for workflow-variables works. When uploading a workflow-variable, we send a capitalized "Content-Type" header. Either the API only processes the capitalized header, or it is broken. I will try fixing the capitalization after posting this feedback (EDIT: this solved the problem).

Lastly, the uninitialized secrets work well! But when you edit an existing secret, we should add an info alert above the editor saying "Existing secrets cannot be retrieved after creating. But you can overwrite the existing value with a new one." When you open a secret for editing now, it just shows an empty editor, which is confusing.

stefan-kracht:
> When I try to save that, it (expectedly) returns a backend error.

Fixed with this behavior. I test the mimetype against our EditorMimeTypeSchema, this test might be a bit too restrictive though:

https://github.com/direktiv/direktiv-ui/assets/121789579/8e55a7e7-d5f9-4005-8bb9-12c5c4be6db8

> I cannot download a binary namespace-variable due to a zod schema error.

I could not reproduce it (I tested it with an image, when I implemented it all the time), for me it works with and without your commit. This might be an OS related issue? However I also applied your fix to the mutation for workflow variables ( c1cb449c80836be5a62815e78e7a88dfc83c2dbd)


> Lastly, the uninitialized secrets work well! But when you edit an existing secret, we should add an info alert above the editor saying "Existing secrets cannot be retrieved after creating. But you can overwrite the existing value with a new one." When you open a secret for editing now, it just shows an empty editor, which is confusing.

Fixed with this behavior:

https://github.com/direktiv/direktiv-ui/assets/121789579/2cceb7ea-3cb3-4e10-8b44-60078d5ebbfb

_video demonstrates how the note will look together with an error, and that it will not show in the create modal or when the secret is not yet initialized_

sebxian (approved):
These changes look good. Approved :+1:

Regarding the secret download error: That's weird. I tried again and this time, I wasn't able to reproduce it either when I went back to the pre-fix version. I am now running he latest stable version of the API.

Maybe it was a one-time glitch. In any case, I would now leave the "Content-Type" headers capitalized. Even if it shouldn't matter, they do work.

PR #474, opened 2023-09-12
[DIR-817] disable the run button when the the editor has unsaved changes

stefan-kracht:
This PR will disable the run button when the editor has unsaved changes.

https://github.com/direktiv/direktiv-ui/assets/121789579/74bdaa68-8d98-4c31-80fe-74a03d88cb19

### The problem
Running a workflow when the user has unsaved changes can create problems. The user may not even be aware that their changes haven't been saved. For example, if the user attempts to save their work but encounters a syntax error, the changes will not be saved at that moment. If the user doesn't notice the error and proceeds to run the workflow, it will execute in an unexpected state. Additionally, all previous changes that were not saved will be lost.

### DIFF
--- a/e2e/explorer/workflow/run.spec.ts
+++ b/e2e/explorer/workflow/run.spec.ts
@@ -167,6 +167,29 @@ test("it is possible to run the workflow by setting an input JSON via the editor
   ).toBe(userInputString);
 });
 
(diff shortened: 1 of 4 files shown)


sebxian (approved):
Very good solution to the problem, I think. Approved :+1:

PR #479, opened 2023-09-14
[DIR-840] add direktiv_api: workflow/v1 to all workflow templates

stefan-kracht:
This PR adds `direktiv_api: workflow/v1` to all of our workflow templates. I created a workflow from each of the templates afterward.


sebxian (approved):
Tested (created and ran workflows from a few of the templates) and approved.

PR #485, opened 2023-09-18
[DIR-728] e2e - test delete namespace

porleaaron:
### Questions

1. How to confirm the namespace is actually gone
2. How to test the navigation after deleting the namespace


stefan-kracht (approved):
Thank you, @porleaaron. I've made some minor changes to your PR and will proceed to merge it. Below is a quick summary of the modifications:

- Optimized the names and descriptions of some tests and test IDs.
- You verified the deletion of the namespace by checking it against the API response. I thought it would be a good idea to perform the same check before deletion to ensure that the namespace appears in the API response.
- The test assumes that other namespaces are already created. While this is likely to be true, I've added an extra namespace in the test setup to ensure that other namespaces are always present.
- Removed a static reference to localhost:3333, which is not valid. The URL and port are set in environment variables, and these variables can change.

Additionally, I've added a second test to cover the scenario where the last namespace is deleted. I wanted to ensure that the user remains on the landing page and is not redirected. This wasn't a requirement when I initially wrote this ticket. I was unsure how to implement it at the time but have since figured out how to mock API responses.

PR #486, opened 2023-09-18
[DIR-773] e2e - JQ Playground

porleaaron:
### What's done
run simple query against a input json

an invalid json will display an error (and the output will be empty)
an invalid query will display an error (and the output will be empty)
the input can be copied via the copy button
the output can be copied via the copy button
test all snippets
clicking a snippet will scroll the user to the top of the page
make sure changing the input will be stored in local storage (test with a reload)
make sure changing the query will be stored in local storage (test with a reload)

### Questions

1. ES lint disabled for one line, need your review on this


stefan-kracht (approved):
Thank you @porleaaron (also for finding the typo I made in the Snippets file :D)  Unfortunately, I changed quite a bit at the end and also added a few tests that you must have forgotten and some that I forgot to include in the ticket. 

As I already told you, I think we should not leak our test logic in the application code, this is why I have removed your changes in the `src/pages/namespace/JqPlayground/Examples/exampleSnippets.ts` and moved that logic into a utils file near to the test file. I've also slightly altered the pattern, successfully resolving the TypeScript error without the need to disable any linting rules.

I found it advantageous to break the code into smaller, more manageable pieces, thereby isolating the tests from one another. This approach enhances the robustness of our testing suite.

When designing tests, it's essential to always consider the objective of the test and the potential failure points. For instance, in one of your tests, you aimed to verify that the user is scrolled back to the top after a snippet is executed:

```typescript
await page.waitForTimeout(1000); // scroll has animation to go to top, so should wait for it
const isScrolledUp = await page.evaluate(() => window.scrollY === 0);
expect(
  isScrolledUp,
  "scroll should be at the top when click on a snippet run"
).toBe(true);
``` 
This code is checking if the scroll position is at the very top. But this test alone does not cover what we actually want to test. To make this a valid test, you want to make sure that the scroll postion is NOT at the top when the user clicks the button followed by the test you have written.

Some other things worth noting:
- I added some test to check:
  - if input copy button works
  - if copy buttons will be disabled when nothing to copy
  - that the input and query are persisted in localStorage
  - that an Error apears when the query is invalid
  - that the Error goes away after typing
- fixed some bugs
  - running a snippet always used the input of the render just before the button was clicked
  - clearing the input, now disables the copy button

PR #488, opened 2023-09-20
[DIR-814] Fix "Open workflow" button on instances page (was bugged in production only)

sebxian:
See DIR-814 for details.

This fixes a bug that only appeared in production, where the "Open workflow" button on the instances detail page didn't work. In development mode, there was no issue with the button.

I removed (commented) the code in src/design/Button/index.tsx that checks whether the child is a Link (`isLinkComponent`). Depending on what we decide, we should delete it completely (or fix it).

In production, this check isn't successful in the instance page for the "Open Workflow" button. I could not debug it completely because the zod error messages are very hard to read.

Playwright tests ran successfully locally so I don't think my changes broke anything (but a few of the tests seem to be a bit flaky).

We should discuss whether we should: (a) use the Button with my fix in the future. This means we have to specify "isAnchor" manually and it isn't computed. (b) fix whatever the underlying cause was for the different behavior in development and production.

I may be missing something obvios, but I would vote (a).

My reasoning for that:

We have 2 instances of the button where it is used with "asChild" and a Link is passed in. The method for calculating "isAnchor" is called very often when it doesn't matter, and it is very hard to debug because zod errors aren't logged in production and if they were, they are always very hard to read so debugging remains difficult. So given the fact that there are only 2 use cases I say this code added complexity isn't worth it and adding "isAnchor" manually in rare cases is totally acceptable.

The changes in `src/util/router/pages.tsx` aren't strictly necessary. I just noticed there was always a `?` at the end of an url even without any urlSearchParams. So I added logic to prevent the ? and made the code a bit easier to follow.

But the extra ? didn't actually cause any problems (as I first suspected).

### DIFF
--- a/src/design/Button/index.tsx
+++ b/src/design/Button/index.tsx
@@ -4,6 +4,7 @@ import { Loader2 } from "lucide-react";
 import { Slot } from "@radix-ui/react-slot";
 import { twMergeClsx } from "~/util/helpers";
 
+// asChild only works with exactly one child, so when asChild is true, we can not have a loading property
 type AsChildOrLoading =
   | {
       loading?: boolean;
--- a/src/design/Button/index.tsx
+++ b/src/design/Button/index.tsx
@@ -3,17 +3,19 @@ import * as React from "react";
 import { Loader2 } from "lucide-react";
 import { Slot } from "@radix-ui/react-slot";
 import { twMergeClsx } from "~/util/helpers";
-import { z } from "zod";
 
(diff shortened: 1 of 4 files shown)


stefan-kracht (approved):
Very good fix. As we just discussed, we can use this new approach on defining a Link button, and in the future we will introduce a separate component for a link that looks like a button. 

When the comment is removed, you can merge this👌

Just for reference: The zod error is saying that `displayName` is `undefined`. It's weird that it only happens on a production build.

PR #495, opened 2023-09-27
[DIR-782] e2e - test api commands button

jamesharrison03:
### All done as described

### DIFF
--- a/e2e/explorer/index.spec.ts
+++ b/e2e/explorer/index.spec.ts
@@ -312,7 +312,7 @@ test("it is not possible to create a workflow when the name already exixts", asy
   await page.getByTestId("new-workflow-name").fill(alreadyExists);
   await page.getByTestId("new-workflow-submit").click();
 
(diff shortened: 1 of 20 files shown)


stefan-kracht:
I just noticed that you used http://localhost:3333 a lot in the e2e test. But the domain and port are part if the env variables and can technically be changed at any time. 

We could use `${env.VITE_E2E_UI_HOST}:${env.VITE_E2E_UI_PORT}` to get that but I would prefer to infer that from the pages url programatically like `new URL(page.url()).origin` if possible. Please check all open PRs (except DIR-726, I already made changes there)

Some other example from the delete workspace tests:

This
```typescript
 expect(page.url(), "url should navigate to the very initial state").toBe(
    "http://localhost:3333/"
  );
``` 
can be written as
```typescript
const frontpageUrl = new URL(page.url());
expect(frontpageUrl.pathname, "url should navigate to the landingpage").toBe(
  "/"
);
```

PR #496, opened 2023-10-02
[DIR-898] update nvm section in README.md

julia-zielke:


stefan-kracht (changes requested):
I think you don't even have to specify the version number and nvm will automatically detect the required version to install.

I just tested it by changing the version in the `.nvmrc` file and run `nvm install`.


Suggested change:

- when the required nvm version is not installed on your machine yet, you have to run `nvm install` once to do so.

stefan-kracht (approved):

PR #497, opened 2023-10-02
Yassir/extend file type

sir-hassan:
## Description

Add type "service" to file types enum.


stefan-kracht:
This is for the new services, right? Do you need this to be merged somewhere to test something? If this is for me, for developing, you can change the target branch to  `stefankracht/dir-899-new-services`. That's where I will work on the new services.

PR #501, opened 2023-10-09
[DIR-717][DIR-944] Notification Bell

julia-zielke:
### Description

A new UI feature called 'NotificationMenu' was added to the Navbar.

A new API called 'namespaceLinting' was created to check whether there are uninitialized secrets, this can be accessed via 'useNamespaceLinting'.


stefan-kracht:
@julia-zielke Here is a small list of the next steps that we just talked about:

- [ ] add `className` to the Notification component to enable at least some primitive styling. Things like padding is very likely to be needed
- [ ] the API of the component could include the popover as well to reduce boilerplate (the component could be used like this`<Notification hasMessage={true}>place message...</Notification>`)
- [ ] the `hasMessage` condition should only affect the rendering of the red dot and not the whole component
- [ ] I was also thinking that we may rename `hasMessage` to `showIndicator` as this name is more clear about what it's actually doing
- [ ] use a button instead of a clickable svg/div
- [ ] add a design to storybook to showcase how an actual notification could look like (it should have a text, maybe a title? and a link)
- [ ] add a design to storybook to showcase how a "no notifications" state could look like
- [ ] add a design to storybook to showcase how a "loading" state could look like (see image below for inspiration)

Some implementation detail in the component (that we can also do together)
- [ ] evaluate the messages from the server, display the message and link to a certain page


---

*example of the loading state of the namespace selector:
CleanShot 2023-10-11 at 16 12 01@2x

julia-zielke:
@stefan-kracht 

Basically done, now testing it.

One issue I couldn't figure out by now:
The PopoverContent is not closing when you click on itself (maybe the <link> needs an additional attribute?).
Do you have an idea how to handle that?

stefan-kracht:
I will look into that. 

I will also change the base branch of this PR to `develop` instead of my branch. This looks very good already and we can make the final touches in this PR

stefan-kracht:
As we just discussed

- [ ] use secrets icon instead of the settings icon
- [ ] introduce a general title that is always visible (the translation file must be updated as well, "Notifications" is currently placed inside "secrets")
- [ ] all design components for notification should be in one file
- [ ] add multiple notifications into the story (there must be some kind of separator between them and it would be nice to see that in a stroy)
- [ ] `useNotificationConfig` could go into a separate file like a `config.ts`


We can do this together
- [ ] tbd: we must trigger a refetch at some point

julia-zielke:
I noticed some things, that I would like to talk about:

- Not sure whether or not the component "NotificationTitle" should be placed inside of Notification. I decided for now to keep it outside, because I wanted to still have the possibility for different Titles on different Scenarios. Also I didn't want to add another child/parameter to the Notification.

- I introduced the component `NotificationMenuSeparator` mainly because the `DropdownMenuSeparator` has a class of "-mx-1", that I don't want to use. Probably there is a better solution? I also would have liked to include a class there with this behavior `:last-of-type:not(:only-of-type):hidden`, but that didn't work as expected.
- 
- The small red dot is a child of the button and gets its hover style through "group-hover", but is there a similar way of Inheritance for this class `data-[state=open]:bg-transparent` on the button?

stefan-kracht:
VERY good work, Julia! The design component is very clean and well-structured. I totally agree with introducing a separate title component, because of all the arguments that you mentioned. Same with the `NotificationMenuSeparator`. Even if it looks very much the same as the separator of the dropdown, I like that it has no dependency to it. 

I will have a look into the other questions that you had.

stefan-kracht:
BTW, this is a very cool feature in our PRs. We automatically generate a preview link, which is super handy for the reviewer to quickly make a functional review and you can also share it with colleagues as a preview or request for feedback.

https://github.com/direktiv/direktiv-ui/assets/121789579/65159995-acbe-4cea-80e0-bd4ece84e4d4

stefan-kracht:
Here is a small checklist from what we have just talked about:

- [ ] `last:hidden` can be removed from separator
- [ ] icon and icon may get`h-4 w-4` 
- [ ] notification title is mistakenly inside the loop instead of outside
- [ ] check for a better way of refetching @stefan-kracht 
- [ ] check if we can change the border of the indicator based on the buttons state @stefan-kracht

stefan-kracht:
> The small red dot is a child of the button and gets its hover style through "group-hover". But is there a similar way of Inheritance for this class data-[state=open]:bg-transparent on the button?
> (This issue is shown in the video below)

FYI: I used `group-data-[state=open]:border-white` to update the border in this case

https://github.com/direktiv/direktiv-ui/assets/121789579/c22eadf6-df8a-4c3b-87f6-8ef3f3bff4c8

stefan-kracht:
@sebxian since @julia-zielke and I worked on this together, I think it would be good if you review this PR. The e2e tests will be done separately to keep the PR minimal. I also think that everything is functional already and is ready to be merged.

There is one thing that I want to point out:

### The strategy on how we trigger a refetch of the notifications
Currently, there is just one notification about secrets that are not yet initialized, but in the future there might be a much longer list. A change in the list of notifications is indirectly invoked by a mutation. 

A) How to get an "uninitialized secrets" notification:
- The user creates a namespace from a git repository that uses some secrets.
- The user creates a workflow with a secret.
- The user updates a workflow by introducing a new secret (or renaming an old one).

B) How to get rid of an "uninitialized secrets" notification:
- Initialize a secret in the secrets.
- Delete the uninitialized secret.

#### How I think it should **not** be done:
In theory, every mutation can trigger a notification as a side effect (like the 5 mutations above), and we should avoid having to inject refetch calls all over the place. We will very likely lose track of all the dependencies and will introduce a lot of complications.

#### How we implemented it right now:
Since the list of notifications is very short, I have implemented a manual refetch for the second list (B) of mutations that I listed above (see 463b26cdc748e502663b0f87fa80c288ae828abd). I think it's very important to refetch the notifications when the user "fixes the issue" to make that notification go away immediately. But for simplicity, I think it's okay to not cover the first list (A) for now and wait for one app reload (or even just a refocusing of the browser window) to see these notifications.

#### How we could do it in the future:
When the complexity grows, I would like to introduce a pattern, where we have one place where we control the refetches. And this should be within the notifications component itself. I'm still investigating on how this can be done technically, but here is my idea in theory:
- For the mutations above, we add a `mutationKey` (see options in https://tanstack.com/query/v4/docs/react/reference/useMutation) as an identifier.
- Within the config of the notifications component, we define which mutationKey(s) belong to which notifications.
- The notification component "somehow" listens to all `onSuccess` callbacks of a mutation.
- When it finds the matching `mutationKey` in the config, it will invalidate the query key for the notifications (which triggers a refetch)

src/componentsNext/NotificationMenu/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+        {isLoading && (
+          <NotificationLoading>
+            {t("components.notificationMenu.isLoading.description")}
+          </NotificationLoading>
+        )}
+        {showIndicator ? (
+          possibleNotifications.map(
+            ([notificationType, notificationConfig], index, srcArr) => {
+              const isLastListItem = index === srcArr.length - 1;
+              const matchingNotification = data.issues.filter(
```
sebxian, line 45:
```suggestion
              const matchingNotifications = data.issues.filter(
```
Arrays should be named in plural.

sebxian (changes requested):
Very nice work, the component looks and works great :rocket:

I have made a few suggestions in the code. One of them is a bit more complex because I wasn't able to understand the code at first and think it should be restructured a bit to make it easier to follow. I hope that comment makes sense, @julia-zielke, feel free to ask me for help or clarifications if needed!

@stefan-kracht , I have two thoughts on how we handle updating the notifications.

1. In my test, after creating a workflow using secrets, I did not immediately see a notification, but would have expected one. I am not sure if that falls under what you wrote about keeping the complexity low, and I agree that in general we have to keep this simple. But in this particular instance, from a UX perspective, it would make sense to refetch the lint results. Please let me know what you think (I am not sure how much complexity this would add).

2. Regarding "How we could do it in the future", I think at some point we should have one general "status" API (using server-side events or WS) that will send status updates about all kinds of issues (namespace linting, service health, errors ...). This would enable us to update the notification bell in real time without refetching, and we wouldn't have to refetch a range of different endpoints.

src/pages/namespace/Settings/Secrets/index.tsx
```diff
@@ -46,6 +48,12 @@ const SecretsList: FC = () => {
 
   const { mutate: deleteSecretMutation } = useDeleteSecret({
     onSuccess: () => {
+      /**
+       * When the user has uninitialized secrets, this will be shown behind
+       * the notification bell. The user might just have deleted that secret
+       * and we need to update the notification bell.
+       */
```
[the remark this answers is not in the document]
stefan-kracht: This one comes from me, so I might answer that. I wanted to explain why exactly we refetch notifications here, because reading the code alone will not give you any hint why that is necessary. Like explained in my comment, I don't like that we have to do refetching here, but as long as we don't have a better solution, I think it is a good idea to exactly explain these notifications side effect.

stefan-kracht:
@sebxian 

> In my test, after creating a workflow using secrets, I did not immediately see a notification, but would have expected one. I am not sure if that falls under what you wrote about keeping the complexity low, and I agree that in general we have to keep this simple. But in this particular instance, from a UX perspective, it would make sense to refetch the lint results. Please let me know what you think (I am not sure how much complexity this would add).

This was indeed intentional. I wanted to keep manual refetching as minimal as possible, and thought that this notification may not be that important to get immediately after creating a workflow. I assumed that forgetting to set secrets is still an exception, and it's okay to wait for one page reload (or window refocus) to get that reminder. But I'm open to change that when you think it's valuable to do so!

src/pages/namespace/Settings/Secrets/index.tsx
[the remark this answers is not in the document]
sebxian: Thanks, I understand now. I was confused by the wording and would suggest changing this to

```suggestion
      /**
       * When the user has uninitialized secrets, this will be reflected in
       * the notification bell. The user might just have deleted that secret
       * and we need to update the notification bell.
       */
```

sebxian:
> This was indeed intentional. I wanted to keep manual refetching as minimal as possible, and thought that this notification may not be that important to get immediately after creating a workflow. I assumed that forgetting to set secrets is still an exception, and it's okay to wait for one page reload (or window refocus) to get that reminder. But I'm open to change that when you think it's valuable to do so!

@stefan-kracht I think this would be valuable. But I would be ok with leaving this out for now and fixing it later -- I don't want to block merging this branch because of it.

src/design/Notification/index.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+export const NotificationHasMessage = () => (
+  <Notification showIndicator={true}>
+    <NotificationTitle>Notifications</NotificationTitle>
+    <NotificationMenuSeparator />
+    <div className="w-full hover:bg-gray-3 dark:hover:bg-gray-dark-3">
+      <NotificationMessage
+        icon={Settings}
+        text="Settings for the current workflow are incomplete."
+      />
```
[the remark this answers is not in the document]
sebxian: Oh, I see. I'm not sure if it's a problem. In general, storybook and our real use don't have to be strictly identical. 

We must assume some devs will just copy and paste this code, but even then, I don't really see a big problem here in case they only want the hover effect (without the popover stuff we use in the real component)?

I am fine with removing or keeping it.

src/pages/namespace/Settings/Secrets/index.tsx
```diff
@@ -196,6 +204,12 @@ const SecretsList: FC = () => {
         <CreateEdit
           item={editItem}
           onSuccess={() => {
+            /**
+             * When the user has uninitialized secrets, this will be shown behind
```
sebxian, line 208:
```suggestion
             * When the user has uninitialized secrets, this will be reflected in
```

stefan-kracht:
> > This was indeed intentional. I wanted to keep manual refetching as minimal as possible, and thought that this notification may not be that important to get immediately after creating a workflow. I assumed that forgetting to set secrets is still an exception, and it's okay to wait for one page reload (or window refocus) to get that reminder. But I'm open to change that when you think it's valuable to do so!
> 
> @stefan-kracht I think this would be valuable. But I would be ok with leaving this out for now and fixing it later -- I don't want to block merging this branch because of it.

It was easy to implement, so I just did that 359d533aa5242ab6e74d8cda20c1942f412bf8e8. I again added a detailed comment here. I don't like these, but as long as this is an unexpected side effect, it should be documented and also acts as a reminder to refactor this when we scale this feature. 

Just FYI: when you remove an uninitialized secret from a workflow, the notification does not go away. I guess this is for performance reasons(?) but I will ask Alan on monday if this is an expected behavior.

src/pages/namespace/Layout.tsx
julia-zielke, line 1: There were errors found - only showing up in the Playwright UI - that say "namespace is undefined". 
This happens because the Navbar is loaded before the namespace is loaded, maybe because it is further up in the DOM or maybe because if no namespace was selected, the redirect to the namespace in local storage happens.  

Basically the same thing like in line 77/78

julia-zielke, line 1: There were errors found - only showing up in the Playwright UI - that say "namespace is undefined".
This happens because the Navbar is loaded before the namespace is loaded, maybe because it is further up in the DOM or maybe because if no namespace was selected, the redirect to the namespace in local storage happens.

Basically the same thing like in line 77/78

julia-zielke (commented):
There were errors found - only showing up in the Playwright UI - that say "namespace is undefined".
This happens because the Navbar is loaded before the namespace is loaded, maybe because it is further up in the DOM or maybe because if no namespace was selected, the redirect to the namespace in local storage happens.

Basically the same thing like in line 77/78

src/pages/namespace/Layout.tsx
julia-zielke, line 1: There were errors found - only showing up in the Playwright UI - that say "namespace is undefined".
This happens because the Navbar is loaded before the namespace is loaded, maybe because it is further up in the DOM or maybe because if no namespace was selected, the redirect to the namespace in local storage happens.

Basically the same thing like in line 77/78

e2e/events/history.spec.ts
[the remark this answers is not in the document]
stefan-kracht: Just pushed a commit fixes the same issue in a different test 558a395b77e55604cd335db92440f99b3b152b2a

stefan-kracht (approved):

stefan-kracht (approved):

PR #505, opened 2023-10-11
[DIR-911] update versions schema

stefan-kracht:
Updates the version schema to only use the fields that we actually need. This will prevent the UI to throw an error when, flow or functions are not set.


sebxian (approved):
Approved :+1:

PR #508, opened 2023-10-16
[DIR-914] no directory hint when deleting a file

stefan-kracht:
Deleting a YAML file shows: "All content of this directory will be deleted as well." This hint should only be visible when deleting a directory.

### DIFF
--- a/src/pages/namespace/Explorer/Tree/Delete.tsx
+++ b/src/pages/namespace/Explorer/Tree/Delete.tsx
@@ -37,8 +37,9 @@ const Delete = ({
         <Trans
           i18nKey="pages.explorer.tree.delete.commonMsg"
           values={{ name: node.name }}
-        />{" "}
-        {t("pages.explorer.tree.delete.directoryMsg")}
+        />
+        {node.type === "directory" &&
+          ` ${t("pages.explorer.tree.delete.directoryMsg")}`}
       </div>
       <DialogFooter>
         <DialogClose asChild>
--- a/e2e/explorer/index.spec.ts
+++ b/e2e/explorer/index.spec.ts
@@ -358,6 +358,15 @@ test(`it is possible to delete a worfklow`, async ({ page }) => {
     .getByTestId("dropdown-trg-node-actions")
     .click();
   await page.getByTestId("node-actions-delete").click();
+
+  await expect(
+    page.getByText(
+      `Are you sure you want to delete ${name}? This cannot be undone.`,
+      {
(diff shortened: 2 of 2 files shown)


src/pages/namespace/Explorer/Tree/Delete.tsx
```diff
@@ -37,8 +37,9 @@ const Delete = ({
         <Trans
           i18nKey="pages.explorer.tree.delete.commonMsg"
           values={{ name: node.name }}
-        />{" "}
-        {t("pages.explorer.tree.delete.directoryMsg")}
+        />
+        {node.type === "directory" &&
+          ` ${t("pages.explorer.tree.delete.directoryMsg")}`}
```
sebxian, line 42: I think this pattern of combining translation strings with a "common" and a case specific message is still a bit too complex and potentially error prone (even if the tests would prevent that).

So I have an alternative suggestion. Change this part of the translation file:
```json
        "delete": {
          "title": "Delete",
          "commonMsg": "Are you sure you want to delete <b>{{name}}</b>? This cannot be undone.",
          "directoryMsg": "All content of this directory will be deleted as well.",
          "cancelBtn": "Cancel",
          "deleteBtn": "Delete"
        },
```

To something where the two use cases are more clearly separated:
```json
        "delete": {
          "title": "Delete",
          "fileMsg": "Are you sure you want to delete <b>{{name}}</b>? This cannot be undone.",
          "directoryMsg": "Are you sure you want to delete <b>{{name}}</b>? All content of this directory will be deleted as well.",
          "cancelBtn": "Cancel",
          "deleteBtn": "Delete"
        },
```
then we could just show one or the other message, which is far easier to understand at a glance.

```jsx
<div className="my-3">
 {node.type === "directory" ?  {t("pages.explorer.tree.delete.directoryMsg")} : {t("pages.explorer.tree.delete.fileMsg")}
</div>
```

stefan-kracht: Very good suggestion. I like that! I don't know if you intentionally removed the "This cannot be undone." part from the directory message, but I also like that change as well. Just to keep the warning message as short as possible, and "All content of this directory will be deleted as well." is warning enough to make the user aware that this is a destructive action that they might reconsider.  327f9a6313fd0b0fe9be39be5b2601b91d39079f

sebxian: That was not intentional, but you're right, it does look better!

PR #511, opened 2023-10-18
v0.8.0 release

stefan-kracht:
This is the PR of the release redesign release. This should not require a proper review anymore, since the release is already done, and just not have been merged into main.

After merging, we can add a tag and GitHub release notes. We can discuss them in the comments here, but I think they can be very minimal, since we basically changed everything :D Maybe we could cherry-pick [some new features](https://github.com/direktiv/direktiv/releases/tag/v0.8.0)?

Just for reference
- the storybook runner is failing because of a bug, that will be fixed in the next release
- the "Node.js CI / check_updates" action will be deleted


sebxian (dismissed):
Looks good to me.

stefan-kracht:
I still needed to do some commits (mostly cherry-picking from `development`) to pass all the tests:
- playwright tests were broken at the time of the release, so I [cherry-picked the fix that I actually did later](https://github.com/direktiv/direktiv-ui/pull/511/commits/8c9751f3eb198bd846a926b084433c947509ece3)
- I found and fixed [this issue](https://github.com/direktiv/direktiv-ui/pull/511/commits/7d08df38903c48bb153a369de818bd6b1b8dd82a). I misinterpreted the dimension of this and thought that it was responsible for a failing test. So it was not mandatory to fix this.
- also I picked this fix from yesterday, because suddenly the other tests failed as well

sebxian (approved):
Looks good!

PR #512, opened 2023-10-18
[DIR-929] fix TableBody import on broadcast page

stefan-kracht:


stefan-kracht:
will be fixed in https://github.com/direktiv/direktiv-ui/pull/509

PR #516, opened 2023-10-25
[DIR-938] notification bell e2e test

julia-zielke:
### Test 1: „Notification Bell has an inactive state by default“

- Checks if the Notification Bell is rendered
- Checks if it shows the correct text for ‚No Notifications’

### Test 2: „Notification Bell updates depending on the count of Notification Messages“

- Creates a new workflow with 2 secrets
- Checks if the Notification Bell is rendered
- Checks it the red Indicator is visible
- Checks if it shows the correct text for ‚Two Notifications’
- Initializes the first secret
- Checks it the red Indicator is visible
- Checks if it shows the correct text for ‚One Notification’
- Initializes the second secret
- Checks it the red Indicator is NOT visible
- Checks if it shows the correct text for ‚No Notifications’


stefan-kracht:
FYI: The `develop` branch is being discontinued, and we're transitioning to use the `main` branch. I've recently merged the latest changes from `develop` into `main` and adjusted the target of this pull request from `develop` to `main`. Since both branches share the same head, this shouldn't have any impact.

Before I did that, I merged the latest `develop` branch back into this and fixed a merge conflict. I expect that everything works the same way as before, but if not please let me know.

e2e/notificationmenu/index.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+  const notificationBell = page.getByTestId("notification-bell").nth(1);
+
+  await expect(
+    notificationBell,
+    "it renders the Notification Bell"
+  ).toBeVisible();
+
+  expect(
+    page.getByTestId("notification-indicator").nth(1),
+    "the red indicator is visible"
```
sebxian, line 66:
```suggestion
    "the indicator for new messages is visible"
```
Just a minor thing, but it's better to describe the semantics of the indicator. The color isn't relevant to the test (and it could change). If you change this, note that there are a multiple occurrences of this.

e2e/notificationmenu/workflow.ts
```diff
@@ -0,0 +1,19 @@
+export const workflowWithSecrets = `direktiv_api: workflow/v1
```
sebxian, line 1: Another very minor suggestion: So far I think we use files named `utils.ts` or `utils/worflow.ts` for stuff that is imported in tests, so we could change this file name to `utils.ts` to be consistent with that.

sebxian (dismissed):
Good work, @julia-zielke !

I have two minor suggestions, but they are rather stylistic. So I am also approving this already.

sebxian (approved):
Thanks for the changes. Approved!

PR #518, opened 2023-11-01
[DIR-939][DIR-DIR-951] Gateway

stefan-kracht:
This PR is introducing a UI for the new gateway feature. The concept changed after a first demo, so I updated the description of this PR

### Other things outside the scope of the gateway:

- Since we now have a lot of different file types, I updated the "New" button in the header and gave it a bit more structure. This button will also be displayed in empty folders.
CleanShot 2023-12-22 at 09 27 41@2x
- [fixed an issue where variables could not be downloaded](https://github.com/direktiv/direktiv-ui/commit/be8cdcfd492c6253b27aea4c01b81f45b793be14). It is not possible that the mime type is undefined when a variable is created with the UI, however variables can also be created programmatically and have no mime type.

### Demo videos

endpoints

https://github.com/direktiv/direktiv-ui/assets/121789579/e1fd0f99-a0db-4c33-b042-47e70bc29975

consumer

https://github.com/direktiv/direktiv-ui/assets/121789579/2e69a5b2-909b-4fcc-8cd8-59a5f14d314c

Follow up tickets
- [E2E tests need to be added](https://linear.app/direktiv/issue/DIR-1040/e2e-tests-for-gateway)
- ["unsaved changes" does not go away](https://linear.app/direktiv/issue/DIR-1041/unsaved-changes-does-not-go-away)


stefan-kracht:
E2E test all pass against ebb3b86a248c196c4ea9eb2026aea7ffd8b8ed99

src/pages/namespace/Explorer/Tree/components/modals/CreateNew/Gateway/Consumer.tsx
```diff
@@ -0,0 +1,133 @@
...
+}) => {
+  const { t } = useTranslation();
+  const namespace = useNamespace();
+  const navigate = useNavigate();
+  const {
+    register,
+    handleSubmit,
+    formState: { isDirty, errors, isValid, isSubmitted },
+  } = useForm<FormInput>({
+    resolver: zodResolver(
```
sebxian, line 46: This would be a bit nicer to read if we declared a `const resolver` separately, instead of nesting it here.

src/pages/namespace/Explorer/Tree/components/modals/CreateNew/Gateway/Route.tsx
```diff
@@ -0,0 +1,133 @@
...
+}) => {
+  const { t } = useTranslation();
+  const namespace = useNamespace();
+  const navigate = useNavigate();
+  const {
+    register,
+    handleSubmit,
+    formState: { isDirty, errors, isValid, isSubmitted },
+  } = useForm<FormInput>({
+    resolver: zodResolver(
```
sebxian, line 46: This would be a bit nicer to read if we declared a `const resolver` separately, instead of nesting it here.

src/pages/namespace/Explorer/components/ArrayInput.tsx
```diff
@@ -0,0 +1,106 @@
+import { FC, useState } from "react";
```
sebxian, line 1: This file is located under `pages/Explorer/etc`, but isn't it generic enough to live in the general (layout) components folder?

(or should we think about a common location for reusable form elements -- in case we don't have this yet?)

This probably applies to other components below as well.

src/pages/namespace/Gateway/Consumers/Table/Row/SecretInput.tsx
```diff
@@ -0,0 +1,32 @@
+import { Eye, EyeOff } from "lucide-react";
```
sebxian, line 1: This could maybe be renamed to "PasswordInput" or "HiddenInput" -- my thought is: If we would later reuse it as a general component, it might be confused with our record type "secret".

It's even used for a "password" right now, so sticking with that term makes sense.

sebxian (commented):
Impressive work, this looks really good (both in code and UI).

I couldn't review each line of code in-depth due to the scope of this PR after my long-ish absence. So I focused on understanding the broad strokes (which are very clear to understand).

I made a few comments/suggestions in the code as part of this PR and also in DIR-943. The latter includes two visual bugs I encountered while testing.

src/componentsNext/FilePicker/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+            <Breadcrumb
+              noArrow
+              onClick={() => {
+                setPath("/");
+              }}
+              className="h-5 hover:underline"
+            >
+              <Home />
+            </Breadcrumb>
+            {segments.map((file) => {
```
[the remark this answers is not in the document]
krachtstefan: Good suggestion, i added this to [DIR-1039](https://linear.app/direktiv/issue/DIR-1039/file-picker-empty-state) as Julia is currently working on this component

src/componentsNext/FilePicker/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+              }}
+              className="h-auto w-full cursor-pointer p-0 font-normal text-gray-11 hover:underline focus:bg-transparent focus:ring-0 focus:ring-transparent focus:ring-offset-0 dark:text-gray-dark-11 dark:focus:bg-transparent"
+            >
+              <FilepickerListItem icon={FolderUp}>..</FilepickerListItem>
+            </div>
+            <FilepickerSeparator />
+          </>
+        )}
+        <FilepickerList>
+          {results.map((file) => {
```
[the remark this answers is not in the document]
stefan-kracht: Also added to [DIR-1039](https://linear.app/direktiv/issue/DIR-1039/file-picker-empty-state)

src/componentsNext/FilePicker/index.tsx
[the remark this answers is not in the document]
stefan-kracht: Good suggestion, i added this to [DIR-1039](https://linear.app/direktiv/issue/DIR-1039/file-picker-empty-state) as Julia is currently working on this component

src/pages/namespace/Gateway/Consumers/Table/Row/SecretInput.tsx
stefan-kracht: Good one,  I will change that!

src/pages/namespace/Explorer/Tree/components/modals/CreateNew/Gateway/Route.tsx
stefan-kracht: I agree! The problem is that it has a dependency to the `t` function to `unallowedNames`. So I think I will make it a separate function/hook then? I think we had this discussion before and decided against it. If you search for `unallowedNames` you will find a lot of similar resolver.

src/pages/namespace/Explorer/components/ArrayInput.tsx
stefan-kracht: I think I just tend to place components to the closest shared root. But I can move it if you think this makes more sense. I would then  move the other component in the same folder as well

sebxian: I'm not familiar enough with these components to be 100% certain that they are generic enough to be used outside of this context. If they are, I think it makes sense to move them.

src/pages/namespace/Explorer/Tree/components/modals/CreateNew/Gateway/Route.tsx
sebxian: What I mean isn't a big change, the const would still be declared in the same scope and have access to all variables. It would just avoid one layer of nesting / indentation.

Not sure but I think I remember a case where the resolver depends on values returned by the useForm() hook but that doesn't seem to be the case here? 

```
  const resolver = zodResolver(
    z.object({
      name: fileNameSchema
        .transform((enteredName) => addYamlFileExtension(enteredName))
        .refine(
          (nameWithExtension) =>
            !(unallowedNames ?? []).some(
              (unallowedName) => unallowedName === nameWithExtension
            ),
          {
            message: t("pages.explorer.tree.newRoute.nameAlreadyExists"),
          }
        ),
      fileContent: z.string(),
    })
  );

  const {
    register,
    handleSubmit,
    formState: { isDirty, errors, isValid, isSubmitted },
  } = useForm<FormInput>({
    resolver,
    defaultValues: {
      fileContent: defaultEndpointFileYaml,
    },
  });
  ```

Comment (stefan-kracht) [reply src/pages/namespace/Explorer/Tree/components/modals/CreateNew/Gateway/Route.tsx:46]: Gotcha. I will resolve it here and make another PR

File: src/pages/namespace/Explorer/components/ArrayInput.tsx (hunk 3 above)
Comment (stefan-kracht) [reply src/pages/namespace/Explorer/components/ArrayInput.tsx:1]: I will do that then.

File: src/pages/namespace/Explorer/Consumer/ConsumerEditor/index.tsx [hunk 7]
```diff
@@ -0,0 +1,118 @@
...
+    const toSave = jsonToYaml(data);
+    updateRoute({
+      path,
+      fileContent: toSave,
+    });
+  };
+
+  return (
+    <Form defaultConfig={consumerConfig}>
+      {({
```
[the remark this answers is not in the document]
sebxian: I agree, let's leave this as is, making an extra component will not improve readability due to more prop drilling.

PR #521, opened 2023-11-27
[DIR-946] Filepicker Component

julia-zielke:
Introducing a design for a new component called Filepicker
The basic functionality is a scrollable overlay with many items that opens upon button click.

### DIFF
--- a/src/design/Filepicker/index.stories.tsx
+++ b/src/design/Filepicker/index.stories.tsx
@@ -103,11 +103,9 @@ export const WithManyItemsBreadcrumbHeadingAndCloseFunctionAtItemClick = () => (
     <FilepickerList>
       {items.map((element) => (
         <Fragment key={element.filename}>
-          <FilepickerClose>
-            <FilepickerListItem icon={element.icon}>
-              {element.filename}
-            </FilepickerListItem>
-          </FilepickerClose>
+          <FilepickerListItem icon={element.icon} asChild>
+            <FilepickerClose>{element.filename}</FilepickerClose>
+          </FilepickerListItem>
           <FilepickerSeparator />
         </Fragment>
       ))}
--- a/src/design/Filepicker/index.stories.tsx
+++ b/src/design/Filepicker/index.stories.tsx
@@ -1,7 +1,17 @@
-import { File, Folder, FolderUp } from "lucide-react";
+import { Breadcrumb, BreadcrumbRoot } from "../Breadcrumbs";
+
+import {
+  File,
+  Folder,
+  FolderUp,
+  LucideIcon,
+  MenuSquareIcon,
+  SettingsIcon,
+} from "lucide-react";
 
 import {
   Filepicker,
+  FilepickerClose,
   FilepickerHeading,
   FilepickerList,
   FilepickerListItem,
@@ -9,6 +19,7 @@ import {
 } from "./";
 
 import type { Meta, StoryObj } from "@storybook/react";
+import { Fragment } from "react";
 
 const meta = {
   title: "Components/Filepicker",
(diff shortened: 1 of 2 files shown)


src/design/Filepicker/index.stories.tsx
```diff
@@ -0,0 +1,105 @@
...
+  <Filepicker>
+    <FilepickerHeading>Collection of Files</FilepickerHeading>
+    <FilepickerSeparator />
+    <div className="w-full hover:bg-gray-3 dark:hover:bg-gray-dark-3">
+      <FilepickerListItem icon={FolderUp}>..</FilepickerListItem>
+    </div>
+    <FilepickerSeparator />
+    <FilepickerList>
+      {items.map((element) => (
+        <div key={element.index}>
```
stefan-kracht, line 96: Can we make this `<Fragment key={element.index}>` or is the `div` needed for CSS?

src/design/Filepicker/index.stories.tsx
```diff
@@ -0,0 +1,105 @@
...
+    <Filepicker {...args}>content goes here...</Filepicker>
+  ),
+  argTypes: {},
+};
+
+export const WithFewItems = () => (
+  <Filepicker>
+    <FilepickerHeading>Collection of Files</FilepickerHeading>
+    <FilepickerSeparator />
+    <div className="w-full hover:bg-gray-3 dark:hover:bg-gray-dark-3">
```
stefan-kracht, line 32: These wrapper divs can be removed, right? In my understanding the `FilepickerListItem` is already rendering that div

src/design/Filepicker/index.stories.tsx
```diff
@@ -0,0 +1,105 @@
...
+  </Filepicker>
+);
+
+type Listitem = {
+  filename: string;
+  index: number;
+};
+
+const items: Listitem[] = [
+  { filename: "image.jpg", index: 0 },
```
stefan-kracht, line 62: This is very minor thing, but I would also add the file icon to that list to add some variation there as well. And since the filename is unique in that array, we could probably remove the `index` attribute and use the filename as a key.

src/design/Filepicker/index.stories.tsx
```diff
@@ -0,0 +1,105 @@
...
+  { filename: "Readme7.txt", index: 17 },
+  { filename: "Readme8.txt", index: 18 },
+  { filename: "Readme9.txt", index: 19 },
+  { filename: "Readme10.txt", index: 20 },
+  { filename: "Readme11.txt", index: 21 },
+];
+
+export const WithManyItems = () => (
+  <Filepicker>
+    <FilepickerHeading>Collection of Files</FilepickerHeading>
```
stefan-kracht, line 88: I think in one of the two stories, we could show how the breadcrumb can be used here. I think this would be a very handy reference to see how we can combine the filepicker with the breadcrumb

src/design/Filepicker/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+      <PopoverContent className=" bg-gray-1 dark:bg-gray-dark-1" align="start">
+        {children}
+      </PopoverContent>
+    </Popover>
+  </div>
+);
+
+export {
+  Filepicker,
+  FilepickerClose,
```
stefan-kracht, line 109: Maybe we can showcase the `FilepickerClose` in one of the stories as well. E.g. clicking a file could close the picker.

src/design/Filepicker/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+  FilepickerHeading,
+  FilepickerList,
+  FilepickerListItem,
+  FilepickerListItemText,
```
stefan-kracht, line 113: As far as I see, we only use `FilepickerListItemText` once. I wonder if we just remove this as a component and directly use the HTML that it produces at this very one place.

stefan-kracht (changes requested):
Very nice work Julia, I like how you have implemented this. I made some comments directly in the code, but only very small suggestions.

src/design/Filepicker/index.tsx
julia-zielke: True, I think that's better.

stefan-kracht (approved):

PR #532, opened 2023-12-12
[DIR-1027] make 403 redirect to logout path

stefan-kracht:
This PR will redirect the user to the logout path, when a fetch request response with a 403 AND the app is running in enterprise mode. This is a fairly easy change in general, but it required some slightly bigger changes in how we deal with end variables (see description below)

I also changed the logout path for dev environment (vite proxy) 

### problems with our environment variables
- I needed to add a condition somewhere in the api factory (in the `getPermissionStatus` method) to redirect the user depending on an environment variable
- environment variables in the Vite app are imported from the `src/config/env/index.ts` file
- it uses `import.meta.env` to access environment variables
- `import.meta.env` can not be processed in Playwright (the runner runs in a node environment and throws an `Cannot use 'import.meta' outside a module` error)
- Playwright uses the api factory as well
- so adding this condition would break playwright

### solution
- we have to make sure that we only use `import.meta.env` in a Vite context, the best way to do that is to only import it in the vite config. 
- In the vite config we [inject a VITE](https://github.com/direktiv/direktiv-ui/pull/532/files#diff-6a3b01ba97829c9566ef2d8dc466ffcffb4bdac08706d3d6319e42e0aa6890dd) variable to the environment
- we [add a global type declaration](https://github.com/direktiv/direktiv-ui/pull/532/files#diff-0acbabdb9e9c1fdb34fb8d527e9db3afa3e5e20435072b18cfed1f60ed4062e5) to tell Typescript that these process.env.VITE might be there and describe how it is shaped.
- From now on, we use env variabls without importing (e.g. `process.env.VITE?.VITE_IS_ENTERPRISE`) but use them as they might be undefined (TS helps with that)


stefan-kracht:
@jensg-st I added you as a reviewer. You can approve when we have tested it on an enterprise build.

sebxian (commented):
The code changes look good to me (but I have been able to test it in the EE context, so Jens should give the final approval).

jensg-st (approved):

PR #534, opened 2023-12-17
[DIR-1022] Filepicker component with state

julia-zielke:
Suggested Setup for the FilePicker Component in our environment:

- Parameter 'onChange' is required to handle the output
- Parameter 'Namespace' is required to know which files to load
- Parameter 'defaultPath' is optional, if empty will be set to root directory "/"
 
...

Updated Setup after Review:
- Parameter 'onChange' is optional to handle the output value = the file that was picked
- Parameter 'Namespace' is optional, if empty it is set by 'useNamespace'
- Parameter 'defaultPath' is optional, if empty it is set to root directory "/"

Example:
` <FilePicker onChange={(newValue) => {console.log(newValue);}} namespace="test" defaultPath="/folder/subfolder"/>`


src/componentsNext/Filepicker/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+  defaultPath?: string;
+  onChange: (newValue: string) => void;
+}) => {
+  const [path, setPath] = useState(defaultPath ? defaultPath : "/");
+
+  const givenNamespace = namespace ? namespace : "";
+
+  const { data, isFetched, isError } = useNodeContent({
+    path,
+    givenNamespace,
```
stefan-kracht, line 36: The `givenNamespace` sounds a little bit strange when reading it here. I would expect it so be just `namespace`. Inside the hook, it makes perfect sense, since there is a variable naming collision. But what we could do is the following:
- calling the property of the parameter object `namespace`
- when we destructure it inside the hook, we rename it to `givenNamespace`
```
  ...
  enabled = true,
  namespace: givenNamespace,
  ...
```

src/componentsNext/Filepicker/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+import { useTranslation } from "react-i18next";
+
+const FilepickerMenu = ({
+  namespace,
+  defaultPath,
+  onChange,
+}: {
+  namespace: string;
+  defaultPath?: string;
+  onChange: (newValue: string) => void;
```
stefan-kracht, line 28: Maybe `newValue` could be named `filePath`? Just to make it a bit more clear, what the string actually contains? 

I would also suggest making the `onChange` optional, even if the user might always need one, I like to have the onChange match the DOM elements behavior. The same way you can also place an input field somewhere and don't do anything with the user's input.

You typed the namespace as required but also implemented a backup if the namespace is `falsy`. I would just make the namespace optional as well

src/componentsNext/Filepicker/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const givenNamespace = namespace ? namespace : "";
+
+  const { data, isFetched, isError } = useNodeContent({
+    path,
+    givenNamespace,
+  });
+
+  // ... Quick Fix - Error Handling needs to be inside useNodeContent, like for the namespace
+  if (isError) setPath("/");
+  // ...
```
stefan-kracht, line 41: If there is an error, you could display a note like "could render a tree, make sure namespace/the-path-to-file exists"

public/locales/en/translation.json
```diff
@@ -1070,6 +1070,12 @@
         }
       }
     },
+    "gatewayForms": {
+      "Filepicker": {
```
stefan-kracht, line 1074: I would remove gatewayForms here, since it can also be used elsewhere and "Filepicker" should be "filepicker"

stefan-kracht (changes requested):
Great work Julia, you made a very nice filepicker!

I converted this to ready to review since we are very close to merge. I have some feedback, you might know about some of the issues since the PR was just a draft, but I will just list everything that I have found to save us another review round

There are two of the same files with different names, you probably just wanted to keep one? I think `FilePicker/index.tsx` is the one with the better name? If yes, the component inside this file should be renamed to the same name


The overlay closes when I enter a folder. I think this could be because of `if (!isFetched) return null;`, but I'm not sure. We could add a spinner when the tree is loading.

https://github.com/direktiv/direktiv-ui/assets/121789579/13124704-6867-4d48-8804-b27a90a59283

It would be nice if we could make it possible to make the whole line clickable like in our file browser

The breadcrumbs height changing depending on if it has content or not

https://github.com/direktiv/direktiv-ui/assets/121789579/bcff5d24-2d99-4d0b-99e7-11e4bd785a24

stefan-kracht (approved):
Good job julia,  

I made some small changes that I hope you are fine with. Please merge if you don't have anything to add.

- removed the demo component (it's better when we don't commit demo code like that. There is a small chance that it will be missed in a review. This may have happened to me 🫠)
- I changed the error to only show on error. When you check for `!data`, it will also show for a blink of a second when the files are loading 
- I gave the file picker a fixed height and width to avoid layout shifts like that: 

https://github.com/direktiv/direktiv-ui/assets/121789579/f2e48314-6023-4676-8552-dfb2bd4cca81

stefan-kracht:
@julia-zielke Great, thanks for cleaning this up. I have to admit I rushed this trough this morning and did not check it on mobile. I also noticed the problem, folder were not full width clickable but wanted to save it as an improvement ticket. Great that you fixed that already.

This looks super nice now, I just tested it within the gateway form. It just works and it looks great!

PR #535, opened 2023-12-19
[DIR-1032] New NamespaceSelector component

julia-zielke:
Introducing a new component called NamespaceSelector, where you can only select, but not search for a namespace or create a new one.

Example:
<NamespaceSelector onValueChange={(value) => {console.log(value);}} />


stefan-kracht (approved):
I just [changed the translation key](https://github.com/direktiv/direktiv-ui/pull/535/commits/ade2c84ca9fad5b6b406824c92cc757ebdd78065) to not include breadcrumb. Other than that: 👌

PR #536, opened 2023-12-21
[DIR-1033] FilePicker - Fix for Scrolling Behavior in A Modal

julia-zielke:
Problem was: 
Vertical scrolling through the files does not work, if the Filepicker is put in a `<Dialog>`


julia-zielke:
Good to know! 😀 

If the <Popover> is put in a <Dialog> you have to add the parameter 'modal' to the Popover Root Element to get focus handling.

`<Popover modal>`

stefan-kracht (approved):
Super nice work Julia for catching this one. I would have probably needed a lot longer to find that one.

- I changed the story a bit to use the design systems button as a dialog trigger. Just for consistency in general.
- also pushed one commit to make all test pass again (I screwed that up in the gateway branch)

Feel free to merge, if you are okay with my changes as well.

PR #539, opened 2024-01-03
Fix filtering by after/before with exact time

sebxian:
The query string created by the UI was wrong (filter.type "MATCH" instead of "AFTER" or "BEFORE" after picking a t@ime).

@stefan-kracht setting `[field]` dynamically in this line of code means TypeScript will not ensure that the following values are valid.  Do you have any ideas on how to improve this?

```
    setFilter({
      [field]: { type: field, value: date },
    });
 ```

### DIFF
--- a/src/pages/namespace/Events/History/components/Filters/RefineTime.tsx
+++ b/src/pages/namespace/Events/History/components/Filters/RefineTime.tsx
@@ -31,7 +31,7 @@ const RefineTime = ({
 
     date.setHours(hr, min, sec);
     setFilter({
-      [field]: { type: "MATCH", value: date },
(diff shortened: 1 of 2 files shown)


stefan-kracht (approved):
Great that you could fix this issue.

> @stefan-kracht setting [field] dynamically in this line of code means TypeScript will not ensure that the following values are valid. Do you have any ideas on how to improve this?

The TS problem, that you described here seems to be harder to fix than I thought. I did not find a good solution to that (without having to refactor) either.

You can merge this if you like. It would be great if you could check, if all e2e tests are still passing locally, because I think we have some e2e test that cover the filters (is it worth to add a simple test that covers this fix as well?)

PR #540, opened 2024-01-04
[DIR-943] postponed review suggestions

sebxian:
## Description

See DIR-943 for discussion.

In some cases, it will be easier to discuss my suggestions as changes in the code (see commit messages for my reasoning).

### DIFF
--- a/src/componentsNext/Breadcrumb/ExplorerBreadcrumb/BreadcrumbSegment.tsx
+++ b/src/componentsNext/Breadcrumb/ExplorerBreadcrumb/BreadcrumbSegment.tsx
@@ -18,15 +18,14 @@ const BreadcrumbSegment: FC<{
    * we need to request file information to figure out which
    * icon to use
    */
-  const requestIconInformation = isLast;
 
   const { data } = useNodeContent({
     path: absolute,
-    enabled: requestIconInformation,
+    enabled: isLast,
   });
 
   if (!namespace) return null;
-  if (requestIconInformation && !data) return null;
+  if (isLast && !data) return null;
 
   const Icon = fileTypeToIcon(data?.node.type ?? "directory");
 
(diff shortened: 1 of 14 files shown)


src/componentsNext/Breadcrumb/ExplorerBreadcrumb/BreadcrumbSegment.tsx
```diff
@@ -18,15 +18,14 @@ const BreadcrumbSegment: FC<{
    * we need to request file information to figure out which
    * icon to use
    */
-  const requestIconInformation = isLast;
 
   const { data } = useNodeContent({
     path: absolute,
-    enabled: requestIconInformation,
+    enabled: isLast,
```
sebxian, line 24: It's easier to follow along without introducing an extra variable (and an extra thought to keep in mind).

src/pages/namespace/Explorer/Workflow/Overview/Services.tsx
```diff
@@ -13,7 +13,7 @@ import { useWorkflowServices } from "~/api/services/query/services";
 const Services = ({ workflow }: { workflow: string }) => {
   const { t } = useTranslation();
   const [dialogOpen, setDialogOpen] = useState(false);
-  const [rebuildService, setRebuildService] = useState<ServiceSchemaType>();
+  const [serviceToRebuild, setServiceToRebuild] = useState<ServiceSchemaType>();
```
sebxian, line 16: When the variable is used further below, it is more obvious than "rebuildService", which at first sight sounds like an actiom/method rather than an object. But I don't insist on changing this. We have similar naming patterns elsewhere (e.g. `deleteRegistry`) so I should probably accept that these variable names are not always 100% obvious.

sebxian:
I rebased my original branch to remove one change, and made a new PR to avoid mix-ups in the git history.

PR #541, opened 2024-01-08
[DIR-912] Switching between views in workflow view wipes content

stefan-kracht:
This PR fixes a bug in the workflow editor where the most recent changes got lost when the user switched to the diagram view and then back to the code editor view.

The bugfix itself was actually very simple, but I had to introduce a `useEffect` as well to handle a side effect that occurs when the user clicks the "revert to previous revision" (luckily, this was covered by a test and was probably the root cause of why the internal state was designed this way). 

I also introduced an e2e to cover the correct behavior.

All e2e test are passing with the api at commit `46e73a0de70f9e0bb75319300933916a08a52517`


sebxian (approved):
Looks good :+1:

PR #542, opened 2024-01-09
[DIR-1039] file picker empty state

julia-zielke:
Changes and new Additions for the FilePicker Component:

- User Feedback through text in the case of an "emptyDirectory" 

- CSS Change to display the container in different heights depending on the number of children element

Refactoring of the Component:
- Instead of the wrapper div surrounding the FilepickerItem, there is now FilepickerButton
- For the Breadcrumbs in the Heading there is now FilepathSegments
- The mapping of the elements in the List is now done in FilepickerItem


julia-zielke:
To Do:
In Storybook there is currently an error when interacting with the Filepicker.
The same error shows at the Popover component and others that are built with Popover like Command.

src/componentsNext/FilePicker/FilepickerItem.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export type FilepickerItemType = {
+  results: NodeSchemaType[];
+  selectable: ((node: NodeSchemaType) => boolean) | undefined;
+  setPath: (path: string) => void;
+  setInputValue: (value: string) => void;
+  onChange: (path: string) => void;
+};
+
+export const FilepickerItem: FC<FilepickerItemType> = ({
```
stefan-kracht, line 20: Should we rename this to `FileList`? My initial thought when reading the name was that this renders one item.

src/componentsNext/FilePicker/FilepickerItem.tsx
```diff
@@ -0,0 +1,66 @@
...
+  FilepickerClose,
+  FilepickerListItem,
+  FilepickerSeparator,
+} from "~/design/Filepicker";
+
+import { NodeSchemaType } from "~/api/tree/schema/node";
+import { fileTypeToIcon } from "~/api/tree/utils";
+import { twMergeClsx } from "~/util/helpers";
+
+export type FilepickerItemType = {
```
stefan-kracht, line 12: For a prop type I think it might be a bit better to end call it `FilepickerItemProps` instead of `FilepickerItemType` (or after the suggested name change below, even `FileListProps`). 

*_I sometimes also end a type name with ...Type, but mostly on a type that I infer from a schema._

src/componentsNext/FilePicker/FilepathSegments.tsx
```diff
@@ -0,0 +1,45 @@
+import { Breadcrumb, BreadcrumbRoot } from "~/design/Breadcrumbs";
+
+import { FC } from "react";
+import { Home } from "lucide-react";
+
+export type FilePathSegmentsType = {
```
stefan-kracht, line 6: As suggested in `src/componentsNext/FilePicker/FilepickerItem.tsx`: We might rename them to `FilePathSegmentsProps`

src/componentsNext/FilePicker/FilepickerItem.tsx
```diff
@@ -0,0 +1,66 @@
...
+                setInputValue(file.path);
+                onChange?.(file.path);
+              }}
+            >
+              <FilepickerListItem icon={fileTypeToIcon(file.type)}>
+                {file.name}
+              </FilepickerListItem>
+            </FilepickerClose>
+          )}
+          <FilepickerSeparator />
```
stefan-kracht, line 61: I think we should hide this `<FilepickerSeparator />` component when it is the last list item, to avoid having two lines?

CleanShot 2024-01-10 at 16 31 54@2x

stefan-kracht (changes requested):
This looks very good, Julia. I think it is well-structured and very readable now! I just a few naming suggestions and one tiny little bug that I commented directly in the code.

src/componentsNext/FilePicker/FilepickerItem.tsx
julia-zielke: Oh... Yes of course!

stefan-kracht (approved):
👌 Approved! Feel free to merge this PR 🚀

PR #544, opened 2024-01-10
[DIR-943] Postponed review fixes

sebxian:
While reviewing [DIR-943](https://linear.app/direktiv/issue/DIR-943) I made some suggestions directly in the code, because it will be easier to just review these changes as a new PR instead of commenting on PRs that are already merged.

These are primarily small typos and similar stuff.


stefan-kracht (approved):
Thanks for the suggestions. They all make perfect sense to me :) 

As just discussed, I will merge right away.

PR #545, opened 2024-01-10
[DIR-1046] add ee gateway plugins

stefan-kracht:
- adds two new plugins that are only available in ee:
  - inbound plugin: event-filter 
  - target plugin: target-event
- It also fixes a problem where a type was not narrowed down correctly (the wrong type did not throw any TS errors until the new plugins were added in this PR)
- The session schema changed in EE and needed to be updated (this only showed a console error, but did not break anything)

It is based on the `stefankracht/dir-1027-make-403-redirect-to-logout-path` branch, as this is the newest gateway branch. For a better diff view, the same branch is set as the target.

### DIFF
--- a/src/pages/namespace/Explorer/Endpoint/EndpointEditor/Form/plugins/Inbound/EventFilterForm.tsx
+++ b/src/pages/namespace/Explorer/Endpoint/EndpointEditor/Form/plugins/Inbound/EventFilterForm.tsx
@@ -17,7 +17,7 @@ import { zodResolver } from "@hookform/resolvers/zod";
 
 type OptionalConfig = Partial<EventFilterFormSchemaType["configuration"]>;
 
-const predfinedConfig: OptionalConfig = {
+const predefinedConfig: OptionalConfig = {
   allow_non_events: false,
 };
 
@@ -44,7 +44,7 @@ export const EventFilterForm: FC<FormProps> = ({
     defaultValues: {
       type: "event-filter",
       configuration: {
-        ...predfinedConfig,
+        ...predefinedConfig,
         ...defaultConfig,
       },
     },
--- a/src/pages/namespace/Explorer/Endpoint/EndpointEditor/Form/plugins/Inbound/EventFilterForm.tsx
+++ b/src/pages/namespace/Explorer/Endpoint/EndpointEditor/Form/plugins/Inbound/EventFilterForm.tsx
@@ -1,0 +1,109 @@
+import { Controller, useForm } from "react-hook-form";
+import {
+  EventFilterFormSchema,
+  EventFilterFormSchemaType,
+} from "../../../schema/plugins/inbound/eventFilter";
+import { FC, FormEvent } from "react";
+import FormErrors, { errorsType } from "~/componentsNext/FormErrors";
+
+import { Card } from "~/design/Card";
+import { Checkbox } from "~/design/Checkbox";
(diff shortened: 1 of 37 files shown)


src/pages/namespace/Explorer/Endpoint/EndpointEditor/schema/plugins/inbound/index.ts
```diff
@@ -2,4 +2,17 @@ export const inboundPluginTypes = {
   acl: "acl",
   jsInbound: "js-inbound",
   requestConvert: "request-convert",
+  eventFilter: "event-filter",
 } as const;
+
+const allPlugins = Object.values(inboundPluginTypes);
+const enterpriseOnlyPlugins = [inboundPluginTypes.eventFilter] as string[];
```
sebxian, line 9: I think this way of flagging EE only plugins isn't ideal. It may be acceptable at this point because we have relatively few plugins.

But what would you think about extending the structure of `targetPluginTypes` to include more information?

Something like:
```
export const targetPluginTypes = {
  instantResponse: { name: "instant-response", enterpriseOnly: false },
  ..
} as const;

I have no idea if this would make things complicated on the typescript front. But in general, I would find it cleaner to have only one list containing all the information.

**Review by sebxian — COMMENTED** [review]
This looks good, just a few minor comments in the code.

File: src/pages/namespace/Explorer/Endpoint/EndpointEditor/Form/plugins/Inbound/EventFilterForm.tsx [hunk 2]
```diff
@@ -0,0 +1,109 @@
...
+import Editor from "~/design/Editor";
+import { Fieldset } from "~/pages/namespace/Explorer/components/Fieldset";
+import { PluginWrapper } from "../components/Modal";
+import { useTheme } from "~/util/store/theme";
+import { useTranslation } from "react-i18next";
+import { zodResolver } from "@hookform/resolvers/zod";
+
+type OptionalConfig = Partial<EventFilterFormSchemaType["configuration"]>;
+
+const predfinedConfig: OptionalConfig = {
```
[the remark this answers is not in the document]
stefan-kracht: I catched them all :D cc00fd9f8c1430d01e2a9896bbacadf2e4460115

src/pages/namespace/Explorer/Endpoint/EndpointEditor/schema/plugins/inbound/index.ts
stefan-kracht: I think this is a very good idea. I had something like this in my mind as well, but did not want to change the structure too much. But I just reviewed it and find it too complicated to read for such a simple filtering, so I did some refactoring like you suggested. It would be great if you can take a final look [at my changes](https://github.com/direktiv/direktiv-ui/compare/cc00fd9f8c1430d01e2a9896bbacadf2e4460115...1c1184b9c6d72e68f0e4a47e9051d885d94e5a78)

stefan-kracht: FYI, target is now `main` since the old target was merged into main `stefankracht/dir-1027-make-403-redirect-to-logout-path `

sebxian (approved):
Thanks for these changes, it looks much better with the extracted filtering method :+1:

Approved!

PR #546, opened 2024-01-10
Fix e2e timing problems

sebxian:
I had several problems with the e2e tests on my laptop. This fixes them (only the diagramEditor tests are still a bit flaky, but I can live with that for now).

As discussed in person, it is probably best to consider my new config values as "safe" and use them as our new default.

It is always a bit hard to be sure with playwright, but I think the main issues were:
- Multiple test runners in parallel were causing lots of timeouts, so I restricted "runners" to 1 in the config.
  - the original behavior can be restored by setting an ENV variable
- Playwright's built in test server needs a timeout to work reliably. I cannot make sense of the value itself (I just set it to the documented default), I only know that there are problems without setting this value here.
- I also set the timeout for expectations to 10s (from default 5s) which helps with a few tests
- I introduced some "waitFor: 'networkidle'" statements.
- Fixed one race condition where I suspect mock data wasn't ready before the request was triggered.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -22,5 +22,8 @@ VITE_IS_ENTERPRISE="FALSE" # true or TRUE will evaluate to true, everything else
 VITE_E2E_UI_HOST="http://localhost"
 VITE_E2E_UI_PORT="3333"
 
+# Enable parallel tests by setting this to "TRUE". Any other value will disable parallel tests.
+VITE_E2E_PARALLEL="FALSE"
+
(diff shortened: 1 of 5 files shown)


stefan-kracht (approved):
Good catch. I'm glad you found a solution for this. It looks good to me and I can confirm the all test are passing with and without running tests in parallel. The total time it took for both test setups to finish did not even differ that much (200 seconds VS 187 seconds).

Feel free to merge 🚀

PR #549, opened 2024-01-11
[DIR-1057] keycloak issue

stefan-kracht:


src/componentsNext/UserMenu/LogoutButton/EnterpriseLogout.tsx
```diff
@@ -0,0 +1,24 @@
+import { DropdownMenuItem } from "~/design/Dropdown";
```
sebxian, line 1: I wonder if it is necessary to have two different components for EnterpriseLogout and OpenSourceLogout.

It seems this is necessary because the authentication methods -- and the respective hooks -- are completely different in the two use cases and there is no other way to call these hooks conditionally. Am I correct? Then I think this is fine.

sebxian (commented):
The code looks good to me!

I left two minor suggestions / questions in the code.

And I think we should make sure that the playwright tests still run against an API that requires an auth token. Have you tried that? I will test it as soon as I figured out how to configure the backend accordingly.

src/componentsNext/UserMenu/LogoutButton/EnterpriseLogout.tsx
stefan-kracht: The reason for that is, that `useAuth();` will throw an error in open source mode (the hook wants to read from a react context, but it does not have a provider) and we can not call a hook conditionally as React does not allow this.

stefan-kracht:
> I think we should make sure that the playwright tests still run against an API that requires an auth token. Have you tried that? I will test it as soon as I figured out how to configure the backend accordingly.


I will try that! I think it works like this:

in the `docker-compose.yaml` in the backend, we have to add this:


```dockerfile
  flow:
    image: ${DIREKTIV_IMAGE}
    command: ["/bin/direktiv", "server"]
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
    environment:
      - DIREKTIV_API_KEY=password    👈
      - DIREKTIV_APP=flow
```

stefan-kracht:
all tests a passing against backend `dee6ec00d79e0bd55c8149b3aa4426404ee21da2`

sebxian (approved):
Thanks for the update & running the tests. Approved!

PR #552, opened 2024-01-12
[DIR-1041] unsaved changes does not go away

stefan-kracht:
My first approach to fix this issues was to utilize `useform` to get a proper state that will tell us if the form has unsaved changes. However, I did not figure out how to do that reliably, so I concentrated on fixing it by comparing the latest server content against current editor state. The challenge (and the reason for this bug) was to compare both of them, even if the order of the lines were different. I solved it by converting them to a JSON and then do a deep equal check (I had to import lodash for that) against each other.


src/pages/namespace/Explorer/utils.ts
```diff
@@ -7,6 +10,27 @@ import { stringify as jsonToPrettyYamlStringify } from "json-to-pretty-yaml";
...
+/**
+ * Compares two yaml strings. Please note that this function will make a
+ * structural comparison, so it will return true if the yaml strings are
+ * structurally equal, even if the order of the keys is different.
+ */
+export const compareYamlStructure = (aYaml: string, bYaml: string) => {
+  const previewJson = yamlToJsonOrNull(aYaml);
+  const fileContentFromServerJson = yamlToJsonOrNull(bYaml);
+  return isEqual(previewJson, fileContentFromServerJson);
+};
```
sebxian, line 32:
```suggestion
export const compareYamlStructure = (aYaml: string, bYaml: string) => {
  return isEqual(yamlToJsonOrNull(aYaml), yamlToJsonOrNull(bYaml));
};
```

Any two files could be compared, so it doesn't make sense in my opinion to create named variables at this point (or if we do, the names should be more generic).

sebxian (commented):
This looks good to me, except for one comment in the code.

src/pages/namespace/Explorer/utils.ts
stefan-kracht: Yes, of course. That was a refactoring artifact. I fixed just fixed that (https://github.com/direktiv/direktiv-ui/pull/552/commits/9c056d091c440a353d21560478d85efe76c708de)

PR #553, opened 2024-01-15
Fix e2e test for usermenu

sebxian:
This fixes the e2e test for the user menu.

It failed on my system and when investigating, I found that no namespace is created before performing this test. So if performed in isolation, the onboarding page will be rendered and the user menu will not be accessible. Probably the test didn't fail in the past because namespaces from other tests existed.


stefan-kracht (approved):
I think I did that on purpose because I thought there is no isolated namespace needed for that test, but I did not think about having no namespaces at all.

PR #554, opened 2024-01-15
rename components and hooks folder

stefan-kracht:
Renaming `componentsNext` to `components` and `hooksNext` to `hooks`. Both folders were created when there was still the old UI in this repository and a `components` and `hooks` folder was already present.

Since we are moving to a monorepo, it is a good time to change the folders now.


sebxian (approved):
E2E tests ran successfully. Approved :rocket:

